Issue #221, opened 2026-05-07
Allow configuration of SoundTouch URLs using port 17000


foob61451:
**Is your feature request related to a problem? Please describe.**

Not all of my SoundTouch speakers were rooted with USB and that increases the time it takes to move to AfterTouch

**Describe the solution you'd like**

Use the service on TCP port 17000 to set the SoundTouch up to use a local AfterTouch instance, for example

```
sys configuration bmxRegistryUrl https://192.168.0.1/bmx/registry/v1/services
sys configuration statsServerUrl https://192.168.0.1/
envswitch boseurls set https://192.168.0.1 https://192.168.0.1/updates/soundtouch
```

**Describe alternatives you've considered**

The USB approach is non-trivial for non technical users. This provides a way to configure a speaker without rooting the speakers.

**Use case**

You could automagically configure any SoundTouch on the network from the AfterTouch web interface using port 17000.

**API Documentation Reference (if applicable)**

The commands are documented here:

https://technical.izndgroup.com/2021/02/connect-bose-soundtouch-10-to-wifi.html

**Priority**
- [ ] Critical - blocks important functionality
- [x ] High - would significantly improve user experience
- [ ] Medium - nice to have enhancement
- [ ] Low - minor improvement

---

**Checklist**
- [ x ] I have searched existing issues to avoid duplicates
- [ x ] I have checked the documentation to ensure this feature doesn't already exist
- [ x ] I have provided a clear use case and rationale
- [ x ] I have considered the impact on existing functionality
- [ x ] I understand this may require SoundTouch API support to implement

gesellix:
I'll have to check, sounds feasible.

Note to self:

```shell
docker run --rm -it --env IP=192.168.123.123 alpine:edge ash -c 'apk add -U busybox-extras && telnet $IP 17000'
```

Maybe even `remote_services on` could work depending on the firmware version.

Other resources:

- https://github.com/sijeffrey/SoundTouch/
- https://www.reddit.com/r/bose/comments/1o5zkym/soundtouch_telnet_probing/

gesellix:
Related: https://github.com/gesellix/Bose-SoundTouch/discussions/219

foob61451:
I tested this with my non-rooted Soundtouch 10 and 20 speakers, to configure them to use your soundtouch-service and it worked.

```
sys configuration bmxRegistryUrl http://192.168.2.1:8000/bmx/registry/v1/services
```
```
[...]

PR #238, opened 2026-05-09
feat: Provide scripts and documentation for on-device install

### LINKED ISSUE #196: Allow 'Serverless' Install on the Speaker itself
**Is your feature request related to a problem? Please describe.**
Currently, AfterTouch constantly needs a server to run on to serve the configuration to the Bose devices. Many people using these devices don't have any server-like hardware in their network and might not be willing to do so.

**Describe the solution you'd like**
I just verified, that the armv7 Version of AfterTouch runs perfectly fine on a SoundTouch 20 itself, without needing any modifications. I have not yet figured out whether and how it could be started in time on boot, as well as how to expose the ports used by aftertouch. A setup like this could either make one of the SoundTouch devices in the network the host for all others or allow a minimalistic 'this speaker only' server.

**Describe alternatives you've considered**
Keeping it the way it is.

**Use case**
Users wouldn't need to have a dedicated server just for AfterTouch in their network.

**SoundTouch API Support**
- [ ] This feature is supported by the official SoundTouch API
- [x] This feature is NOT supported by the SoundTouch API (custom enhancement)
- [ ] I'm not sure if this is supported by the SoundTouch API

**API Documentation Reference (if applicable)**
n.a.

**Implementation Details (optional)**
Described above.

**Device Compatibility**
- SoundTouch 20, so probably all devices, though SoundTouch portable might struggle in terms of performance.

**Examples**
Described above.

**Priority**
- [ ] Critical - blocks important functionality
- [x] High - would significantly improve user experience (due to upcoming EOL)
- [ ] Medium - nice to have enhancement
- [ ] Low - minor improvement

**Additional context**
None.

**Related Issues**
Talked about this a bit on https://github.com/gesellix/Bose-SoundTouch/issues/185

---

**Checklist**
- [x] I have searched existing issues to avoid duplicates
- [x] I have checked the documentation to ensure this feature doesn't already exist
- [x] I have provided a clear use case and rationale
- [x] I have considered the impact on existing functionality
- [x] I understand this may require SoundTouch API support to implement

timvahlbrock:
## Description

Brief description of the changes in this PR.

## Type of Change

Please check the type of change your PR introduces:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)
- [ ] Test improvements
- [ ] Build/CI improvements

## Related Issues

- Fixes #196

### API Changes
- [ ] Added new endpoints
- [ ] Modified existing endpoints
- [ ] Added new CLI commands
- [ ] Modified existing CLI commands
- [ ] Added new configuration options

### Implementation Details

Added script files and documentation on how to install aftertouch on the SoundTouch device.

### Automated Tests
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] All existing tests pass
- [ ] Test coverage maintained or improved

### Manual Testing
- [ ] Tested with real SoundTouch device(s)
- [ ] Tested CLI changes manually
- [ ] Tested in different network environments

**Device(s) tested with:**
- Device model: [e.g. SoundTouch 10]
- Device IP: [e.g. 192.168.1.100]
- Test results: [brief description]

### Test Commands
```bash
# Commands used to test this change
make test
go test ./pkg/client -v -run TestNewFeature
soundtouch-cli --host 192.168.1.100 new-command
```

## Documentation

- [x] Updated relevant documentation
- [ ] Added code comments for complex logic
- [ ] Updated CLI help text
- [ ] Added usage examples
- [ ] Updated API documentation

**Documentation files updated:**
- [ ] README.md
- [ ] docs/API-Endpoints-Overview.md
- [ ] docs/CLI-REFERENCE.md
- [ ] Code documentation (godoc)

## Backward Compatibility

- [x] This change is backward compatible
- [ ] This change includes breaking changes (requires major version bump)
- [ ] This change requires configuration migration

## Security Considerations

- [x] No security implications
- [ ] Security review required
- [ ] Added input validation
- [ ] Updated authentication/authorization

## Performance Impact

- [x] No performance impact
- [ ] Performance improvement
- [ ] Potential performance regression (justify why)

## Code Quality

- [x] Code follows project style guidelines
- [x] No linting errors
- [x] No security warnings
- [x] Memory leaks checked (if applicable)

### Pre-submission Checklist

- [ ] `make check` passes (format, lint, vet)
- [ ] `make test` passes
- [ ] No TODO comments left in production code
- [ ] Error handling is comprehensive
- [ ] Logging is appropriate (not too verbose, not too quiet)

## Additional Notes

Any additional information that reviewers should know:
See issue.

## Review Requests

**Areas that need special attention:**
- [ ] Error handling logic
- [ ] Performance critical sections
- [ ] Security implications
- [ ] API design choices
- [x] Documentation clarity

**Specific questions for reviewers:**
1. Should this be mentioned in the installation instructions?
2. Should we create a follow up issue for the missing port exposure? 

---

**Reviewer Guidelines:**
- Check that all tests pass
- Verify documentation is updated
- Test manually if device access available  
- Consider backward compatibility
- Evaluate error handling and edge cases

### DIFF
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -40,7 +40,7 @@ To update AfterTouch, simply run the installation command again. The installer w
 
 ## Uninstallation
 
-To uninstall AfterTouch, run the following command on the speaker.
+Before uninstall, you might want to revert the migration, especially the changes to the server URLs (even though having configured an unresponsive local server probably is about as bad as having configured unresponsive Bose servers). To uninstall AfterTouch, run the following command on the speaker.
 
 ```bash
 curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/uninstall.sh | sh
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -1,0 +1,48 @@
+# On-Device Installer
+
+Allows to run AfterTouch on SoundTouch devices directly, eliminating the need to run and maintain a separate server on the local network.
+
+
+## Disclaimer
+
+### Invasiveness
+
+AfterTouch usually normally migrates the SoundTouch devices very noninvasive, by changing the configuration of the device. Running AfterTouch on the device itself is slightly more invasive, because it needs to create a script that starts AfterTouch on boot.
+
+### AfterTouch Availability
+
+Some devices will expose the AfterTouch port, some won't. We currently suspect that the newer generation devices (those with Bluetooth) will expose the port, while the older ones won't. We're still investigating how to expose AfterTouch on all devices. 
+
+If your device doesn't expose the port, you can still use the on-device installer, but you'll need to run AfterTouch on each one of your speakers individually and may only access AfterTouch via ssh port forwarding. This will also make OAuth authentication a little more tricky, but should also work via SSH port forwarding.
+
+## Installation
+
+Enable SSH on your SoundTouch device using the usual "Stick with remote_services" method. Connect with the following command.
+
+```bash
+ssh -oHostKeyAlgorithms=+ssh-rsa root@<IP_ADDRESS_OF_SPEAKER>
+```
+
+Then, run the following command to install AfterTouch on the device.
+
+```bash 
+curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/install.sh | sh
+```
+
+After the installation check if you can access AfterTouch from your local device by navigating to `http://<IP_ADDRESS_OF_SPEAKER>:8000`. If you can access the AfterTouch UI, you're good to go! If not, you may need to run AfterTouch on the speaker via SSH port forwarding.
+
+```bash
+ssh -L 8000:localhost:8000 root@<IP_ADDRESS_OF_SPEAKER>
+```
+
+## Updating AfterTouch
+
+To update AfterTouch, simply run the installation command again. The installer will check if there's a new version available and update it if necessary.
+
+## Uninstallation
+
+To uninstall AfterTouch, run the following command on the speaker.
+
+```bash
+curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/uninstall.sh | sh
+```
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -14,6 +14,12 @@ Some devices will expose the AfterTouch port, some won't. We currently (May 2026
 
 If your device doesn't expose the port, you can still use the on-device installer, but you'll need to run AfterTouch on each one of your speakers individually and may only access AfterTouch via ssh port forwarding. This will also make OAuth authentication a little more tricky, but should also work via SSH port forwarding.
 
+### Space Limitation
+
+The storage space on the SoundTouch devices is very limited. At the moment only one AfterTouch installation barely fits on them with enough room for the data it needs to maintain. When installing, make sure that you have removed any binaries and folders of previous installation attempts.
+
+The space limitation also means we are currently unsure on how to update the system, because two binaries are already too large. We are currently working on this - both by checking how we can make the binaries smaller, but also on how we can extend the storage space (e.g. by running AfterTouch from a USB drive).
+
 ## Installation
 
 Enable SSH on your SoundTouch device using the usual "Stick with remote_services" method. Connect with the following command.
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -26,7 +26,7 @@ ssh -oHostKeyAlgorithms=+ssh-rsa root@<IP_ADDRESS_OF_SPEAKER>
 Then, run the following command to install AfterTouch on the device.
 
 ```bash 
-curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/install.sh | sh
+rw && curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/install.sh | sh
 ```
 
 After the installation check if you can access AfterTouch from your local device by navigating to `http://<IP_ADDRESS_OF_SPEAKER>:8000`. If you can access the AfterTouch UI, you're good to go! If not, you may need to run AfterTouch on the speaker via SSH port forwarding.
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -2,7 +2,6 @@
 
 Allows to run AfterTouch on SoundTouch devices directly, eliminating the need to run and maintain a separate server on the local network.
 
-
 ## Disclaimer
 
 ### Invasiveness
--- a/scripts/on-device-install/README.md
+++ b/scripts/on-device-install/README.md
@@ -10,7 +10,7 @@ AfterTouch usually normally migrates the SoundTouch devices very noninvasive, by
 
 ### AfterTouch Availability
 
-Some devices will expose the AfterTouch port, some won't. We currently suspect that the newer generation devices (those with Bluetooth) will expose the port, while the older ones won't. We're still investigating how to expose AfterTouch on all devices. 
+Some devices will expose the AfterTouch port, some won't. We currently (May 2026) suspect that the newer generation devices (those with Bluetooth) will expose the port, while the older ones won't. We're still investigating how to expose AfterTouch on all devices. 
 
 If your device doesn't expose the port, you can still use the on-device installer, but you'll need to run AfterTouch on each one of your speakers individually and may only access AfterTouch via ssh port forwarding. This will also make OAuth authentication a little more tricky, but should also work via SSH port forwarding.
 
--- a/scripts/on-device-install/aftertouch
+++ b/scripts/on-device-install/aftertouch
@@ -1,0 +1,105 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          soundtouch-service
+# Required-Start:    $network $local_fs
+# Required-Stop:     $network $local_fs
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: Run AfterTouch on this device
+# Description:       Start/stop AfterTouch soundtouch-service
+### END INIT INFO
+
+
+NAME="aftertouch-service"
+DESC="Bose AfterTouch service"
+DAEMON="/opt/aftertouch/aftertouch-service"
+PIDFILE="/var/run/$NAME.pid"
+DATADIR="/opt/aftertouch/data"
+SCRIPTNAME="/etc/init.d/$NAME"
+USER="root"
+
+
+# Export PATH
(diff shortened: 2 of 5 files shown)


timvahlbrock:
Just realized that the default version can only be 0.72.0 or later, because the scripts will not exist in earlier versions. Will fix this later today.

gesellix:
> Just realized that the default version can only be 0.72.0 or later, because the scripts will not exist in earlier versions. Will fix this later today.

Yes, makes sense, please use 0.73.0 👍

scripts/on-device-install/aftertouch
```diff
@@ -0,0 +1,105 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          soundtouch-service
```
gesellix:
```suggestion
# Provides:          aftertouch-service
```

scripts/on-device-install/aftertouch
```diff
@@ -0,0 +1,105 @@
...
+# Required-Start:    $network $local_fs
+# Required-Stop:     $network $local_fs
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: Run AfterTouch on this device
+# Description:       Start/stop AfterTouch soundtouch-service
```
gesellix, line 9: I think it's fine to have some place where both names appear. Let's keep it like this.

scripts/on-device-install/aftertouch
```diff
@@ -0,0 +1,105 @@
...
+    mkdir -p "$DATADIR"
+
+    start-stop-daemon --start \
+        --quiet \
+        --pidfile "$PIDFILE" \
+        --background \
+        --make-pidfile \
+        --chuid "$USER" \
+        --startas "/bin/sh" \
+        -- -c "\"$DAEMON\" --data-dir '$DATADIR' --record-interactions=false"
```
gesellix: I recently added an option `--discovery-enabled=false` to disable periodic device discovery. We might add it here, too.

scripts/on-device-install/install.sh
```diff
@@ -0,0 +1,40 @@
...
+update-rc.d aftertouch defaults
+
+echo "Installation complete. Running initial startup to accelerate future startups..."
+/etc/init.d/aftertouch start
+
+/etc/init.d/aftertouch status
+
+echo "Installation complete. Aftertouch $VERSION is now running on your device."
+echo "You can try to connect to at http://<your-device-ip>:8000 ."
+echo "If the connection fails, reconnect ssh with port forwarding like:"
```
gesellix, line 43: 😻

scripts/on-device-install/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Disclaimer
+
+### Invasiveness
+
+AfterTouch usually normally migrates the SoundTouch devices very noninvasive, by changing the configuration of the device. Running AfterTouch on the device itself is slightly more invasive, because it needs to create a script that starts AfterTouch on boot.
+
+### AfterTouch Availability
+
+Some devices will expose the AfterTouch port, some won't. We currently suspect that the newer generation devices (those with Bluetooth) will expose the port, while the older ones won't. We're still investigating how to expose AfterTouch on all devices.
```
gesellix:
```suggestion
Some devices will expose the AfterTouch port, some won't. We currently (May 2026) suspect that the newer generation devices (those with Bluetooth) will expose the port, while the older ones won't. We're still investigating how to expose AfterTouch on all devices. 
```

scripts/on-device-install/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+After the installation check if you can access AfterTouch from your local device by navigating to `http://<IP_ADDRESS_OF_SPEAKER>:8000`. If you can access the AfterTouch UI, you're good to go! If not, you may need to run AfterTouch on the speaker via SSH port forwarding.
+
+```bash
+ssh -L 8000:localhost:8000 root@<IP_ADDRESS_OF_SPEAKER>
+```
+
+## Updating AfterTouch
+
+To update AfterTouch, simply run the installation command again. The installer will check if there's a new version available and update it if necessary.
```
gesellix, line 45: Is this correct? I didn't see logic to fetch the latest version. The user would have to pass the version as command line argument.

scripts/on-device-install/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Uninstallation
+
+To uninstall AfterTouch, run the following command on the speaker.
```
gesellix: Should we mention that uninstalling the service without reverting the migration will put the speaker into invalid configuration?

gesellix (commented):
I left some suggestions and ideas. Overall this looks great.

scripts/on-device-install/aftertouch
timvahlbrock: You mean to avoid stressing out the device? Could we still trigger it from the UI then? Maybe it would be good to run it once on startup and then allow rescanning through the UI? Either way, I think setting it to a rather high value should be enough, I think you had 60m or so In one example. In all the trials I did the device was perfectly fine with discovery enabled. Of course this only really helps if the port is accessible from the outside.

scripts/on-device-install/README.md
timvahlbrock: well, this was assuming that the version in the script is updated. I though about wether the install script could be included in the release notes itself, so it would contain the corresponding release version. But then the user would still have to input the target version on installation 🤔

scripts/on-device-install/README.md
timvahlbrock: oh yeah. probably a good idea.

timvahlbrock: On the other hand, the original servers also don't work anymore anyway, right? Nevertheless it's probably better to leave the device in a clean configuration.

timvahlbrock: I added a note. For now also tried to make clear that this might not have any immediate benefit. Correct me if I'm wrong.

scripts/on-device-install/README.md
timvahlbrock: so  I just checked how much space is left on a device that already runs aftertouch. it seems like there isn't enough space to have two binaries present. So currently an update needs to be done in place (delete the old binary, install the new one). I'm going to check if this is possible with the current scripts.

scripts/on-device-install/README.md
```diff
@@ -0,0 +1,47 @@
...
+Enable SSH on your SoundTouch device using the usual "Stick with remote_services" method. Connect with the following command.
+
+```bash
+ssh -oHostKeyAlgorithms=+ssh-rsa root@<IP_ADDRESS_OF_SPEAKER>
+```
+
+Then, run the following command to install AfterTouch on the device.
+
+```bash 
+rw && curl -sSL https://raw.githubusercontent.com/gesellix/Bose-SoundTouch/main/scripts/on-device-install/install.sh | sh
```
timvahlbrock, line 34: I also noticed I'm using the main branch here in contrast to using a specific version in the install script. So we should either determine wether we want the user to use the main branch for installations or a specific version for everything.

scripts/on-device-install/README.md
timvahlbrock: So it seems like the device has some weird behavior regarding storage space. I can download a 12MB binary on a system that has 8MB space left (maybe the device is cleaning up some things?). As soon as aftertouch runs however deleting said binary does not free up enough space to download another one.

scripts/on-device-install/aftertouch
gesellix: Yeah, my thoughts were about possible performance gains. I'm fine with keeping it as-is, but we might set the discovery interval to something like 60m. Re-scanning every 5m is probably a bad default, we should change that, too, but that doesn't belong into this pull request.

scripts/on-device-install/README.md
gesellix: Let's document those findings in the README. I think we can continue with your pull request and merge it do main. The update mechanism can be improved/fixed in a follow-up.
About the default version in those scripts: I didn't think about that, but you're right, that's also something we should fix. Stale/old versions simply don't make sense.

scripts/on-device-install/README.md
gesellix: Yes, I prefer to clean up after myself, even if it's back to a broken default ;)

scripts/on-device-install/README.md
gesellix: Let's keep using the main branch. Otherwise we would frist have to detect which version is the latest one.
A clean approach would be a release concept for the installer scripts, just like for the other artifacts, but I think that would be a bit of overengineering right now. We should gather feedback from users and then decide if this has to be improved.

scripts/on-device-install/README.md
gesellix: Your note looks good to me.

gesellix:
So, I'm just before merging this. You might address the findings about the [speaker's disk space](https://github.com/gesellix/Bose-SoundTouch/pull/238#discussion_r3214557328) in the README right here or later in another pull request. Just leave a note if I should wait with the merge :)

scripts/on-device-install/README.md
timvahlbrock: Added notes to the README. Also added a variable that determines where the binary for the update will be downloaded. Even if that currently is in a volatile directory or on the USB stick the storage does not seem to be sufficient to do the update. 

Maybe we can also later modify the script so that it can launch AfterTouch from the USB stick and we wouldn't have to be so careful with the storage at all. This would either mean mounting USB earlier or deferring the startup of aftertouch until the system itself mounts the stick. This way, the file system of the device also wouldn't have to be writable the whole time.

gesellix: Yes, sounds good!

scripts/on-device-install/aftertouch
timvahlbrock: Changed to 60m.

gesellix (approved):
LGTM
Thanks a lot for providing the polished scripts!

BramScheepers:
I was literally just about to start installation on a dedicated server when I stumbled upon this PR 😆

timvahlbrock:
> I was literally just about to start installation on a dedicated server when I stumbled upon this PR 😆

That's what I call timing 😂

gesellix:
> I was literally just about to start installation on a dedicated server when I stumbled upon this PR 😆

Fresh release incoming in a vew minutes! 😆 
Please report any findings or success to https://github.com/gesellix/Bose-SoundTouch/issues/221! ... ah. wait. Only in case you would use the telnet approach.

BramScheepers:
> > I was literally just about to start installation on a dedicated server when I stumbled upon this PR 😆
> 
> Fresh release incoming in a vew minutes! 😆 Please report any findings or success to #221! ... ah. wait. Only in case you would use the telnet approach.

Where would you like me to report any other issues with specifically this install method?

gesellix:
> > > I was literally just about to start installation on a dedicated server when I stumbled upon this PR 😆
> > 
> > 
> > Fresh release incoming in a vew minutes! 😆 Please report any findings or success to #221! ... ah. wait. Only in case you would use the telnet approach.
> 
> Where would you like me to report any other issues with specifically this install method?

You already created another issue, that's fine :)
The original issue was https://github.com/gesellix/Bose-SoundTouch/issues/196

PR #292, opened 2026-05-15
Fix internet radio playback on bose soundtouch 10

GameTec-live:
## Description

I wanted to setup this project for my bose soundtouch 10 and while everything seemed to connect correctly, the internet radio playback, my main usecase, just didnt work.

This PR should fix playback (did so in my testing) and implements the needed endpoint(s).

Full transparency: This PR is 100% AI

## Type of Change

Please check the type of change your PR introduces:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)
- [ ] Test improvements
- [ ] Build/CI improvements

### API Changes
- [ x ] Added new endpoints
- [ x ] Modified existing endpoints
- [ ] Added new CLI commands
- [ ] Modified existing CLI commands
- [ ] Added new configuration options

### Implementation Details
- Describe the main changes
- List any new dependencies
- Mention any architectural changes

### Automated Tests
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] All existing tests pass
- [ ] Test coverage maintained or improved

### Manual Testing
- [ ] Tested with real SoundTouch device(s)
- [ ] Tested CLI changes manually
- [ ] Tested in different network environments

**Device(s) tested with:**
- Device model: [e.g. SoundTouch 10]
- Device IP: [e.g. 192.168.1.100]
- Test results: [brief description]

### Test Commands
```bash
# Commands used to test this change
make test
go test ./pkg/client -v -run TestNewFeature
soundtouch-cli --host 192.168.1.100 new-command
```

## Documentation

- [ ] Updated relevant documentation
- [ ] Added code comments for complex logic
- [ ] Updated CLI help text
- [ ] Added usage examples
- [ ] Updated API documentation

**Documentation files updated:**
- [ ] README.md
- [ ] docs/API-Endpoints-Overview.md
- [ ] docs/CLI-REFERENCE.md
- [ ] Code documentation (godoc)

## Backward Compatibility

- [ ] This change is backward compatible
- [ ] This change includes breaking changes (requires major version bump)
- [ ] This change requires configuration migration

**Breaking changes (if any):**
- Describe what breaks
- Provide migration instructions

## Security Considerations

- [ ] No security implications
- [ ] Security review required
- [ ] Added input validation
[...]

PR #249, opened 2026-05-10
Add new stream format in TuneIn query

tony-lopes:
## Description

Update TuneIn query to add stream format "hls" (m3u)

## Type of Change

Please check the type of change your PR introduces:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring (no functional changes)
- [ ] Test improvements
- [ ] Build/CI improvements

## Related Issues

NA

### API Changes
- [ ] Added new endpoints
- [x] Modified existing endpoints
- [ ] Added new CLI commands
- [ ] Modified existing CLI commands
- [ ] Added new configuration options

### Implementation Details

Add new value (hls) to the query parameter format

### Automated Tests
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] All existing tests pass
- [ ] Test coverage maintained or improved

### Manual Testing
- [x] Tested with real SoundTouch device(s)
- [ ] Tested CLI changes manually
- [ ] Tested in different network environments

**Device(s) tested with:**
- Device model: SoundTouch 10
- Test results: The speaker successfully played the TuneIn station "/v1/playback/station/s335608"

### Test Commands
```bash
# Commands used to test this change
make test
go test ./pkg/client -v -run TestNewFeature
soundtouch-cli --host 192.168.1.100 new-command
```

## Documentation

- [ ] Updated relevant documentation
- [ ] Added code comments for complex logic
- [ ] Updated CLI help text
- [ ] Added usage examples
- [ ] Updated API documentation

**Documentation files updated:**
- [ ] README.md
- [ ] docs/API-Endpoints-Overview.md
- [ ] docs/CLI-REFERENCE.md
- [ ] Code documentation (godoc)

## Backward Compatibility

- [ ] This change is backward compatible
- [ ] This change includes breaking changes (requires major version bump)
- [ ] This change requires configuration migration

**Breaking changes (if any):**
- Describe what breaks
- Provide migration instructions

## Security Considerations

- [ ] No security implications
- [ ] Security review required
- [ ] Added input validation
- [ ] Updated authentication/authorization

## Performance Impact

- [ ] No performance impact
- [ ] Performance improvement
- [ ] Potential performance regression (justify why)

**Performance notes:**
- Measured impact: [benchmarks, timing, memory usage]
[...]

PR #300, opened 2026-05-16
feat(bmx): make TuneIn formats= configurable via Settings.TuneInStreamFormats

gesellix:
PR #249 added "hls" unconditionally to TuneIn's Tune.ashx formats= query. That regressed playback on the SoundTouch line: TuneIn returns an .m3u8 HLS playlist for stations like K-LOVE (s33828), the speaker can't parse it, blinks amber and falls silent. Verified that firmware 27 on ST10 and ST20 ships the byte-identical Mozilla CCADB bundle and validates the actual stream chain cleanly, so it isn't a cert-expiry issue (#292's hypothesis). The speaker simply has no HLS support.

Changes:

  - TuneInStream is now a builder, not a const: takes the station ID plus a formats string (empty falls back to the new exported DefaultTuneInStreamFormats = "mp3,aac,ogg", matches the pre-#249 request shape).
  - TuneInPlayback and TuneInPlaybackPodcast take the formats string.
  - New Settings.TuneInStreamFormats string. Empty by default. Operators with HLS-capable speakers can set it to "mp3,aac,ogg,hls", or any other comma-separated list, via settings.json. The value is passed through verbatim; AfterTouch does not validate the individual format tokens, so this is also the right knob for trialling additional formats without code changes.
  - Two regression tests pin both the empty-uses-default contract and the override-passes-through contract (with the whitespace- trim sub-case) so PR #249-style regressions surface at compile/test time.

The setting is settings.json-only (matches the existing pattern for AllowInsecureUpstreamTLS / TrustForwardedHeaders / TrustedProxyCIDRs which are also edit-the-file settings). UI surface can be a small follow-up if reporters ask for it.

Example settings.json snippet to re-enable HLS (only if your speaker can actually play it):

```json
    {
      "server_url": "http://aftertouch.local:8000",
      "tunein_stream_formats": "mp3,aac,ogg,hls"
    }
```

Restart soundtouch-service after editing.

Related to https://github.com/gesellix/Bose-SoundTouch/issues/292.

### DIFF
--- a/pkg/service/bmx/bmx.go
+++ b/pkg/service/bmx/bmx.go
@@ -20,10 +20,34 @@ import (
 // TuneIn endpoint templates used to resolve station and stream URLs.
 const (
 	TuneInDescribe     = "https://opml.radiotime.com/describe.ashx?id=%s"
-	TuneInStream       = "http://opml.radiotime.com/Tune.ashx?id=%s&formats=mp3,aac,ogg,hls"
 	TuneInNavigateAshx = "http://opml.radiotime.com/?render=json"
 	TuneInSearchAPI    = "https://api.radiotime.com/profiles?fulltextsearch=true&version=1.3&query="
+
+	// DefaultTuneInStreamFormats is the comma-separated format list
+	// AfterTouch sends to TuneIn's Tune.ashx by default. Matches the
+	// pre-2026-05-10 behaviour from before PR #249 added "hls"
+	// unconditionally — HLS playback is broken on SoundTouch 10/
+	// firmware 27 (and probably the rest of the line; see #292).
+	// Speakers receive an .m3u8 playlist URL they can't parse, blink
+	// amber, fall silent. Operators with HLS-compatible speakers can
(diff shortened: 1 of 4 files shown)


gmuth:
I am not sure what the issue here is/was but wanted to report that my ST20 Series I (using soundploy) still plays HLS:

```xml
<ContentItem source="LOCAL_INTERNET_RADIO" type="stationurl"
   location="http://soundploy.gmuth.de/klove.json" />
```

```json
{
  "audio": {
    "hasPlaylist": true,
    "isRealtime": true,
    "streamUrl": "https://maestro.emfcdn.com/stream_for/k-love/tunein/hls"
  },
  "name": "K-LOVE",
  "streamType": "liveRadio"
}
```

The test case has been created by resolving the tunein station via the usual bmx endpoint. I've removed some optional attributes for simplification and used LOCAL_INTERNET_RADIO instead of TUNEIN to make it work independent from the configured bmx registry and services.

PR #348, opened 2026-05-20
fix: removing stale devices from datastore

### LINKED ISSUE #346: Dropdown entries not taking into account name changes
**Describe the bug**
The dropdown menus at "Data Sync" and "Migration" don't show new names when they are changed. The "Devices" tab does show them

Image

**To Reproduce**
1. Migrate a device
2. Change its name using the Bose Soundtouch App
3. Rescan. Device name will update in Devices tab
4. Look at Migration or Data Sync. It wont be updated there.

**Expected behavior**
Should update.

@gesellix unless you are already aware of this and working on it, you can just assign me

mennemann:
## Description

Fixes #346.

I am not entirely sure if we should merge this as is... The thing that confuses me, is that `/setup/devices` still returns the old name, while `/setup/info/<id>` returns the new one.

### DIFF
--- a/pkg/service/bmx/radiobrowser.go
+++ b/pkg/service/bmx/radiobrowser.go
@@ -20,6 +20,7 @@ func RadioBrowserSearch(query string) (*models.BmxNavResponse, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	defer func() { _ = resp.Body.Close() }()
 
 	if resp.StatusCode != http.StatusOK {
--- a/pkg/service/datastore/datastore.go
+++ b/pkg/service/datastore/datastore.go
@@ -1412,6 +1412,24 @@ func (ds *DataStore) RemoveDeviceDir(account, device string) error {
 	return ds.RemoveDevice(account, device)
 }
 
+// MoveDevice atomically moves a device directory from one account to another
+// on the same filesystem. If the target account directory doesn't exist it is
+// created. Returns an error if the rename fails (leaving the source intact).
+func (ds *DataStore) MoveDevice(oldAccount, newAccount, deviceID string) error {
+	ds.fileMutex.Lock()
+	defer ds.fileMutex.Unlock()
+
+	oldDir := ds.AccountDeviceDir(oldAccount, deviceID)
+	newDir := ds.AccountDeviceDir(newAccount, deviceID)
+
+	newAccountDir := filepath.Dir(newDir)
+	if err := ds.rootMkdirAll(newAccountDir, 0755); err != nil {
+		return err
+	}
+
+	return ds.rootRename(oldDir, newDir)
+}
+
 // DeduceSourceIDs updates the source IDs in the given slice by deducing them from recents and presets.
 func (ds *DataStore) DeduceSourceIDs(account, device string, sources []models.ConfiguredSource) {
 	// Deduce source IDs from recents and presets
--- a/pkg/service/datastore/datastore_test.go
+++ b/pkg/service/datastore/datastore_test.go
@@ -457,3 +457,40 @@ func TestSettingsPersistence(t *testing.T) {
 		t.Errorf("Expected DiscoveryEnabled %v, got %v", settings.DiscoveryEnabled, loaded.DiscoveryEnabled)
 	}
 }
+
+func TestMoveDeviceMigratesData(t *testing.T) {
+	tempDir := t.TempDir()
+	ds := NewDataStore(tempDir)
+
+	oldAccount := "default"
+	newAccount := "8637922"
+	deviceID := "F4E11E930BEB"
+
+	// Seed the device under the old account with presets
+	presets := []models.ServicePreset{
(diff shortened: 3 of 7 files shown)


mennemann:
Okay, the issue is a bit deeper than I thought:

When a device is first discovered without an account id (e. g. after factory reset), its entry is written to `data/accounts/default/devices/<device_id>`. If I then migrate the device and set an account id, it will be rediscovered with that account id and therefore be stored at `data/accounts/<account_id>/devices/<device_id>`. Everything that gets discovered lands in the correct location, however the UI always refers to the one in `data/accounts/default/...`. 

This causes (among potentially other things):

- The device names in drop-downs and the Account tab are wrong. 
- The device is shown under `default` in the Account tab.
- Even after Data Sync, the Presets arent shown in the Account tab (because they are written to the one with the correct account id, but read from default). 


This can all be fixed by copying Presets, Sources, and Recents and deleting the device under the old account_id when we discover a device with the same device id but different account id.

gesellix:
Thanks @mennemann, your follow-up comment nails the diagnosis precisely. A few notes on how this lands against what shipped in [v0.90.0](https://github.com/gesellix/Bose-SoundTouch/releases/tag/v0.90.0), plus a couple of safety concerns.

## Where this fits

v0.90.0 already covers part of the territory, with operator-confirmed gates:

- The `presets_recents_sources_consistency` check (from v0.88, branch `i343-consistency`) detects exactly this case, a device exists under both `default` and a real account, and offers `FixIDDeleteOrphanAccountEntry` as a one-click QuickFix on the Health tab. It deletes the stale entry; it doesn't preserve data.
- The new `default_account_non_bose_devices` check + Evict QuickFix in v0.90.0 covers a different case (non-Bose UPnP devices that answered the generic `MediaRenderer:1` SSDP probe). Plus the UPnP manufacturer filter at discovery time keeps them from being persisted in the first place.

Your PR adds **automatic** handling for the legitimate-Bose-speaker-now-migrated case, instead of requiring the operator to find the Health-tab finding and click. That's the right level of automation here: the speaker explicitly tells us its current account via `MargeAccountUUID` in `/info`, so the move is unambiguous and matches the speaker-wins principle we apply elsewhere. Worth saying so in the PR description so future readers know which of the three mechanisms addresses which case.

As a side-effect, the trigger condition (`live != "" && stored != "" && live != stored`) also covers real-account-A → real-account-B transitions, not just `default` → real-account. That's actually useful but unmentioned in the description. We should try to be clear about that to help users being aware of that behaviour.

## Concerns

1. **Data-loss hazard (narrow but real).** The three saves use `_ = s.ds.SaveX(...)`, so any error is silently discarded, and we still proceed to `RemoveDevice(storedAccount, deviceID)`. If disk-full or a permission error hits between the first save and the delete, we lose the original. The pre-migration `default/` dir is usually almost empty in practice, so the realistic risk is low, but the code shape is risky enough that I'd like to tighten it. See the linter checks for some hints.

2. **Hard-coded list of files.** Today we copy Presets, Recents, ConfiguredSources. We don't copy `DeviceInfo.xml` (probably fine, the new discovery already writes a fresh one) or anything else that lives per-device. If a future commit adds a new per-device file, this code silently won't migrate it. A `DataStore.MoveDevice(oldAccount, newAccount, deviceID)` that does a single directory rename (`os.Rename`) on the device dir would be atomic, future-proof, and one line at the call site.

3. **CI red checks.** The Lint check is failing on this PR. The other red checks (Vulnerability Scan, Static Security Analysis, CodeQL, Security Summary) are pre-existing on every PR in this repo and unrelated to your change. I usually run the linter like this `golangci-lint run --fix` to directly apply obvious fixes.

4. **No tests.** A small unit test exercising the `default → realAccount` transition (with non-empty presets under `default/` to assert the data was preserved) would lock the behaviour in. Happy to add it after the `MoveDevice` refactor... I'll let the AI help with that anyway.

## Suggested next step

Refactor the three-saves-and-delete block into a single `s.ds.MoveDevice(storedAccount, accountID, deviceID)` call, returning an error if the rename fails. Then the call site can keep its existing position. I can take the `MoveDevice` part as a follow-up commit on this branch if that's easier; just say the word.

In any case, thanks again! Your diagnosis was clearly written and easy to follow, which is more than half the work.

mennemann:
Should be done now, let me know you think anything else needs work on this :)

gesellix (approved):

gesellix (approved):
Thanks for iterating on this so quickly! All four original concerns are addressed — LGTM.

- **`355acc7`** — initial fix (superseded by next commit).
- **`db9834f`** — `MoveDevice` replaces copy+delete: directory-level `rename(2)` covers every file (no hard-coded list), and leaves the source intact if it fails (no silent data loss). Test added. Minor nit: doc says "atomically" but it's a `mkdir` + `rename` pair under `fileMutex` — "leaves source intact on failure" is more precise. Not blocking.
- **`3e66788`** — lint. `make test` clean.

**On the `/setup/devices` vs `/setup/info/<id>` discrepancy**: not a regression. `/setup/info` fetches live from the speaker; `/setup/devices` reads from disk and updates on the next discovery scan. A Rescan after the rename is enough to reconcile them — no code change needed.

Issue #515, opened 2026-06-19
[Compatibility]: SoundTouch Portable (412540), FW 27.0.6 — USB remote_services AND #471 telnet envswitch trick both fail to bring up sshd


Henri-be:
### Speaker model

SoundTouch Portable, Series I, Model 412540

### Firmware version

27.0.6.46330.5043500 epdbuild.trunk.hepdswbld04.2022-08-04T11:20:29

### What works?

Sorry I don't speek very well english and information below are from Claude (AI) further our tests.

Discovery / `/info` over HTTP :8090
`setup inspect --telnet` — full identity, network, sources, presets read correctly via port 17000
Telnet port 17000 reachable and responsive (`->` banner, accepts commands)
`envswitch boseurls set "<marge>" "<swUpdate>"` — accepted, returns `OK`, and is persisted (confirmed via `getpdo CurrentSystemConfiguration` echoing back exactly what was set)
`sys configuration swUpdateUrl <url>` — accepted, returns `OK`

### What doesn't work?

USB `remote_services` method: created an empty `remote_services` file on a freshly FAT32-formatted USB stick (via Rufus, non-bootable, default cluster size), inserted into the speaker's USB port while powered via mains (not battery), full power-cycle (AUX button for 10 seconds) with the stick already inserted. Port 22 remains `Connection refused` every time. This matches the firmware era noted in the project's own TELNET-COMMAND-REFERENCE.md: "The Portable on more recent firmware drops further commands and is the hardest target."
`setup enable-ssh` (#471/#476 port-17000 envswitch trick): ran via `soundtouch-cli --host <ip> setup enable-ssh`. The command completes successfully — telnet banner received, `envswitch boseurls set "https://aftertouch.invalid;touch /tmp/remote_services;/etc/init.d/sshd start" "https://aftertouch.invalid/update"` returns `OK`. `WaitForSSHPort` times out after 90s as expected per the documented "some devices only start sshd after a restart" caveat.
Followed the suggested next step (power-cycle, then retry): power-cycled the speaker (AUX button for 10 seconds) multiple times after the injection. Port 22 stays `Connection refused`.
[...]

Issue #471, opened 2026-06-07
Enable remote services using 17000 envswitch boseurls


foob61451:
In a previous issue there were changes to set the URLs using the 17000 port: https://github.com/gesellix/Bose-SoundTouch/issues/221

But this does not provide a way to set the CA Certificate if using a self-signed untrusted CA on the soundtouch-service HTTPS port.

There is a way to enable remote services using 17000, even for newer firmwares using this command:

```
telnet speaker 17000
...
envswitch boseurls set "https://server;touch /tmp/remote_services;/etc/init.d/sshd start"  "https://server/update"
```

These commands will execute when the marge URL is next checked, which can be up to ~60 seconds.

The device can then be accessed using SSH:

```
sleep 65
ssh -o "HostKeyAlgorithms=+ssh-rsa,ssh-dss" root@speaker
```

Once this is confirmed, use `envswitch boseurls set ` on port 17000 to set the URLs back to the normal soundtouch-service URLs, e.g.:

```
envswitch boseurls set "https://server" "https://server/update"
```

Now that the speaker is accessible, we can persist access, secure the root account, and disable access to 17000.

Remount the root fs as read-write so we can make changes:
```
mount / -o rw,remount
```
Change root password:
```
passwd root
<new pass>
<new pass>
```
Set SSH key (optional):
```
mkdir -m 700 /home/root/.ssh
echo "ssh-ed25519 ...." > /home/root/.ssh/authorized_keys
```
Persist remote access:
```
cat << "EOF" > /mnt/nv/rc.local
#!/bin/bash
touch /tmp/remote_services
/etc/init.d/sshd start
EOF
chmod 755 /mnt/nv/rc.local
```
Close the 17000 port by adding a rule to /etc/init.d/Firewalls/update_iptables at the end:
```
cat << "EOF" >> /etc/init.d/Firewalls/update_iptables

# Block 17000
iptables -I INPUT -p tcp --dport 17000 -j DROP
iptables -I INPUT -p tcp --dport 17000 -i lo -j ACCEPT
EOF
```

Now reboot the speaker. The SSH port should be accessible with root's password (or private key), and the 17000 port should no longer be accessible to the network. 

The existing code to migrate the device using SSH and set the CA certificate can now be used.

gesellix:
```
envswitch boseurls set "https://server;touch /tmp/remote_services;/etc/init.d/sshd start"  "https://server/update"
```

[...]

PR #519, opened 2026-06-22
setup enable-ssh: add --full-config for devices where sshd never starts (#515)

gesellix:
## Why

The default `setup enable-ssh` (#471) injects the `remote_services` / `sshd start` payload only through `envswitch boseurls set` and relies on the speaker re-reading its boseurls (~60s) without a reboot. On the **SoundTouch Portable (Series I, model 412540, FW `27.0.6.46330.5043500`)** (issue #515) and on some **CineMate 520** units the device accepts and persists that injection (`getpdo` confirms the value) but `sshd` never comes up, so `:22` stays "Connection refused".

@Henri-be got root on the ST Portable by typing a different sequence by hand over telnet :17000:

- the injection rides `sys configuration margeServerUrl` (the runtime layer) **as well as** `envswitch`,
- all four URL keys are written,
- the device is **rebooted** so it re-parses the config at boot.

## What

Add an opt-in `--full-config` flag to `setup enable-ssh` that replicates that exact sequence (`EnableSSHViaTelnetFullConfig` + telnet reboot via the existing `RebootMethodTelnet`):

```
sys configuration bmxRegistryUrl "<base>/bmx/registry/v1/services"
sys configuration statsServerUrl "<base>"
sys configuration margeServerUrl "<base>;touch /tmp/remote_services;/etc/init.d/sshd start"
sys configuration swUpdateUrl    "<base>/updates/soundtouch"
envswitch boseurls set "<base>;touch /tmp/remote_services;/etc/init.d/sshd start" "<base>/updates/soundtouch"
getpdo CurrentSystemConfiguration
sys reboot
```

The default single-`envswitch` path is **unchanged**, so the field-confirmed flow on the Wireless Link Adapter and the CineMate 520 `lisa` variant does not regress.

## Tests / docs

- New unit test asserts the exact six-command sequence (injection present in both `margeServerUrl` and `envswitch`).
- `make check` / `golangci-lint` clean.
- Docs updated to describe both paths and which models/firmware need `--full-config`: `TELNET-COMMAND-REFERENCE.md` (new "What we use to enable SSH" section) and `DEVICE-LOGGING.md` (CLI method added to "Enabling Root Access").

## Status

The `--full-config` automation is **candidate behaviour awaiting reporter confirmation**: the manual sequence is confirmed working on the ST Portable, but the flag that automates it has not yet been re-confirmed on hardware. Some units (certain ST10 / CineMate 520 firmwares) respond to neither path and need the serial / U-Boot console route instead.

Refs #515.

### DIFF
--- a/cmd/soundtouch-cli/cmd_setup.go
+++ b/cmd/soundtouch-cli/cmd_setup.go
@@ -538,6 +538,53 @@ func setupSSHCheckCmd() *cli.Command {
 	}
 }
 
+// runEnableSSHInjection runs the port-17000 SSH-enable injection over telnet,
+// printing the device transcript as it goes. With fullConfig it sends the
+// #515 sequence (all four config URLs with the injection on margeServerUrl, not
+// just envswitch) and reboots afterwards; otherwise it sends the single-
+// envswitch default that fires on the speaker's next boseurls check.
+func runEnableSSHInjection(m *setup.Manager, host, serviceURL string, fullConfig bool) error {
+	var (
+		logs string
+		err  error
+	)
+
+	if fullConfig {
+		fmt.Printf("Enabling SSH on %s via telnet :17000 (full #515 sequence: all four config URLs with the injection on margeServerUrl, then reboot)...\n", host)
+		logs, err = m.EnableSSHViaTelnetFullConfig(host, serviceURL)
+	} else {
+		fmt.Printf("Enabling SSH on %s via telnet :17000 (runs on the speaker's next boseurls check, up to ~60s)...\n", host)
+		logs, err = m.EnableSSHViaTelnet(host, serviceURL)
+	}
+
+	if logs != "" {
+		fmt.Print(logs)
+	}
+
+	if err != nil {
+		PrintError(err.Error())
+		return err
+	}
+
+	if !fullConfig {
+		return nil
+	}
+
+	fmt.Println("Rebooting the speaker to apply the new configuration...")
+
+	rlogs, rerr := m.Reboot(host, setup.RebootMethodTelnet)
+	if rlogs != "" {
+		fmt.Print(rlogs)
+	}
+
+	if rerr != nil {
+		PrintError(rerr.Error())
+		return rerr
+	}
+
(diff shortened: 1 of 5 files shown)


Henri-be:
Hello,
I'm back at my primary residence and am now trying to migrate my three speakers: ST20, ST300, and ST Portable.
First test to enable SSH on my ST20. I’m trying --full-config switch :

soundtouch-cli --host 192.168.1.6 --full-config setup enable-ssh

No success.

The configuration change was applied successfully:

> → getpdo CurrentSystemConfiguration
margeServerUrl {
  text: "https://aftertouch.invalid;touch /tmp/remote_services;/etc/init.d/sshd start"
}
statsServerUrl {
  text: "https://aftertouch.invalid"
}
swUpdateUrl {
  text: "https://aftertouch.invalid/updates/soundtouch"
}
isZeroconfEnabled {
  text: true
}
usePandoraProductionServer {
  text: true
}
saveMargeCustomerReport {
  text: false
}
bmxRegistryUrl {
  text: "https://aftertouch.invalid/bmx/registry/v1/services"
}

->OK

I suspect that a valid URL is required for this to work. When it worked for me on my ST Portable, I had temporarily installed Aftertouch on my laptop and configured it to point to it.


Translated with DeepL.com (free version)

Henri-be:
I tried using various valid URLs, but without success.
Could it be that an Aftertouch server needs to be installed?
That would be really unfortunate for those who want to install Aftertouch directly on the speaker.

gesellix:
> Could it be that an Aftertouch server needs to be installed?

I don't think so, because that example with "aftertouch.invalid" is only required to look like a valid url, so that the commands after the semicolon `;` would be executed.
I'm unsure, though, if we're missing some detail.

Please let's continue the thread at https://github.com/gesellix/Bose-SoundTouch/pull/519 so that the discussion won't get lost in this closed pull request.

gesellix:
> No success.

Please provide some more details what exactly didn't work. I assume you mean the SSH not being available even after reboot?

Please leave your feedback at #471.

PR #644, opened 2026-08-23
Retry configured player devices during discovery

Mr-Tao:
## Summary

- Route `soundtouch-player --devices` entries through unified discovery's configured-device path.
- Retry configured speakers whenever discovery runs instead of trying them only once during process startup.
- Preserve mDNS and UPnP discovery before probing configured hosts, so an unreachable configured speaker cannot consume the network discovery timeout.
- Correct the stale `-host` example in the player README to use `--devices`.

## Linked issue

No linked issue. This is a small, focused fix for a hardware-reproduced startup timing bug.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

- [ ] `make check` passes (the Docker-based HTTP-client integration suite was not available locally)
- [ ] Tested against a real SoundTouch device (the bug and restart workaround were reproduced on hardware; the patched binary has not been deployed there)
- [x] `go test -count=1 ./...`
- [x] `go test -race -count=1 ./...`
- [x] `go vet ./...`
- [x] `golangci-lint v2.13.1 run`
- [x] Regression test verifies that a configured speaker unavailable during the first discovery is added after it becomes reachable and discovery runs again.

## Reproduction

Start `soundtouch-player` with a speaker in `--devices` while that speaker is offline. Bring the speaker online and trigger Discover in the web UI. Before this change the speaker remains absent because the manual host is tried only by the startup goroutine; restarting the player makes it appear. With this change, each discovery run includes the configured host and retries its `/info` probe.

## Checklist

OpenAI Codex assisted with implementation, tests, and review. I reviewed the complete diff and ran the validation listed above.

### DIFF
--- a/cmd/soundtouch-player/README.md
+++ b/cmd/soundtouch-player/README.md
@@ -88,7 +88,7 @@ go build -o soundtouch-player
 ./soundtouch-player -port 8888
 
 # Connect to specific device
-./soundtouch-player -host 192.0.2.100
+./soundtouch-player --devices 192.0.2.100
 ```
 
 ### Command Line Options
@@ -137,7 +137,7 @@ device datastore).
 The application automatically discovers SoundTouch devices using:
 - **mDNS discovery** for local network devices
 - **UPnP/SSDP discovery** as fallback
-- **Manual device addition** via IP address
+- **Configured devices** via `--devices`, retried whenever discovery runs
 
 ### Real-time Updates
 The interface maintains WebSocket connections to each device for instant updates of:
--- a/cmd/soundtouch-player/main.go
+++ b/cmd/soundtouch-player/main.go
@@ -170,6 +170,15 @@ func main() {
 
 				webApp.BroadcastDiscoveryStatus("starting", webApp.DeviceCount())
 
+				// Register configured devices immediately rather than waiting for
+				// the full mDNS/UPnP sweep below (bounded by cfg.DiscoveryTimeout,
+				// currently 10s) to complete. manualHosts are also folded into
+				// discoveryService's PreferredDevices so a host that's offline
+				// right now still gets retried on every subsequent discovery pass.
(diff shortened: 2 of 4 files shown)


gesellix:
Thanks for this, it's a nice fix for a real gap (a `--devices` host that's
offline at startup was gone for good until a restart).

I went through it and found a few small things worth tightening up. Rather
than list them all here, I put them into a stacked PR on top of your
branch: https://github.com/Mr-Tao/Bose-SoundTouch/pull/1

Short version of what it changes:

- `NewDiscoveryService` could add the same host to `PreferredDevices`
  twice if it was already configured via `PREFERRED_DEVICES`; added a
  dedup check.
- `classifySource` compared `DiscoveryMethod` with `==`, but it can be a
  `+`-joined composite (e.g. `Configuration+mDNS/Bonjour`) when a
  configured host is also found via mDNS/UPnP in the same sweep — that
  case fell through to "discovered". Switched to a substring match.
- `main.go`'s startup goroutine dropped the immediate registration of
  `--devices` hosts (it now only registers them through the first
  discovery sweep, delayed by up to 10s); restored the immediate
  registration loop alongside the new retry-on-sweep behavior, so both
  work together.
- Added a couple of tests, including one adopting `httptest.NewTestServer`
  (Go 1.27).

I tested all three fixes on real hardware (mixed `--devices` +
`PREFERRED_DEVICES` config, an offline-then-online retry, and a device
discovered both manually and via UPnP in the same sweep) and they all
behave as expected.

Once CI is green on the stacked PR I'll merge it in, then merge this one.

gesellix:
I just noticed that I can push my follow-up commits directly to your branch, so I got rid of the stacked pr.
image

gesellix:
Thanks @Mr-Tao, your fix will be included with the next release!

PR #647, opened 2026-08-23
setup: add telnet URL rollback

Mr-Tao:
## Summary

Add an explicit telnet-only rollback path for speakers migrated through port 17000. The CLI, setup API, and web UI can restore the four canonical Bose service URLs without requiring SSH or an XML backup.

The rollback uses the same ordered runtime writes, `envswitch` persistence, and exact `getpdo` readback as telnet migration. This change also:

- rejects command-unsafe URL input before opening a speaker connection;
- rejects telnet-only URL overrides with the SSH rollback method;
- serializes URL-changing telnet sequences per speaker within one service process, including telnet reboot operations;
- offers rollback for any live non-canonical URL set, including configuration left by an older AfterTouch backend;
- accepts both persistence acknowledgements observed on SoundTouch firmware, while rejecting bare, reordered, duplicated, or otherwise ambiguous replies;
- reports partial runtime and uncertain persistence outcomes without claiming an atomic rollback.

The operation intentionally restores only the four URL fields. It does not change SSH persistence, account pairing, DNS, CA, presets, or filesystem state, and the speaker still needs a reboot followed by verification.

## Linked issue

Refs #221

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

Completed on the amended branch:

- `go test ./cmd/soundtouch-player ./cmd/soundtouch-service ./pkg/service/setup ./pkg/telnet`
- `go test -race ./pkg/service/setup ./pkg/telnet`
- `go vet ./pkg/telnet ./pkg/service/setup ./pkg/service/handlers ./cmd/soundtouch-cli ./cmd/soundtouch-player ./cmd/soundtouch-service`
- `git diff --check`

A full `go test ./...` run reached one unrelated live TuneIn integration test failure caused by the external service response; all packages affected by this change passed. The complete GitHub CI gate should rerun after the amended commit is pushed.

Real-device validation used a backed-up SoundTouch 30 in standby and outside a zone. Telnet migration to an AfterTouch endpoint and exact-override rollback to the canonical Bose URLs both returned success. After each operation, the speaker was rebooted and all four URL fields matched the requested values. Volume, zone membership, and standby state were preserved.

The device exposed a second valid `envswitch` response shape:

```text
Setting Bose Server URLs to <marge> and <update>
->OK
->
```

The parser now accepts exactly that shape and the previously covered single-line prompt-suffix shape. Focused tests reject bare confirmation, bare `OK`, reversed or duplicated lines, and unexpected annotations.

The canary separately confirmed that inventory import is not preset/data sync. That migration preflight concern is outside this URL rollback and should be handled by a focused follow-up rather than expanding this PR.

## Checklist

### DIFF
--- a/cmd/soundtouch-cli/cmd_setup.go
+++ b/cmd/soundtouch-cli/cmd_setup.go
@@ -1510,12 +1510,23 @@ func renderMigrationSummary(deviceIP, serviceURL string, s *setup.MigrationSumma
 	}
 }
 
-// setupRevertCmd wraps setup.Manager.RevertMigration — the same operation
-// as the web UI's "Revert to Defaults" button (Migrate tab). Restores
-// SoundTouchSdkPrivateCfg.xml, /etc/hosts, and /etc/resolv.conf from their
-// .original backups, removes the AfterTouch DNS-hook artifacts, and strips
-// just the AfterTouch-labeled cert out of the trust bundle. No --service-url
-// needed: everything it touches already lives on the speaker.
+var telnetRevertOverrideFlags = []string{"marge-url", "stats-url", "sw-update-url", "bmx-url"}
+
+func validateRevertMethodOptions(method string, overrideFlags []string) error {
+	if method != "ssh" && method != string(setup.MigrationMethodTelnet) {
+		return fmt.Errorf("unsupported revert method %q; expected ssh or telnet", method)
+	}
+
+	if method != string(setup.MigrationMethodTelnet) && len(overrideFlags) > 0 {
+		return fmt.Errorf("--%s requires --method telnet", strings.Join(overrideFlags, ", --"))
+	}
+
+	return nil
+}
+
+// setupRevertCmd restores either the SSH/filesystem migration state or only
+// the four URL fields written by a telnet migration. The default remains the
+// existing SSH path for backwards compatibility.
 //
 // Deliberately out of scope (matches the web UI button): SSH/remote_services
 // persistence (use `setup remote-services --remove`) and account pairing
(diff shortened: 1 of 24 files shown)


Mr-Tao:
I amended the branch after validating it on a real SoundTouch 30. That firmware returns the `envswitch` confirmation as a message followed by `->OK`; the previous parser reported HTTP 500 even though all four writes had succeeded.

The amended parser accepts exactly the two response shapes now observed on hardware and rejects ambiguous variants. With that fix, both migration to an AfterTouch endpoint and exact-override rollback to the canonical Bose URLs returned success and survived a reboot with all four fields verified. Volume, zone membership, and standby state were preserved.

The focused normal and race tests plus `go vet` pass. I also updated the PR body with the exact test coverage and the separate data-sync caveat found during the canary.

gesellix:
Thanks for this, and for the hardware validation notes. I have rebased the
branch onto `main` and pushed three fixes on top. Two things happened to your
own commit in the process, so they are worth stating plainly before the rest.

## Your commit was rewritten

The rebase was not a fast-forward: `main` moved under this branch while
[PR #657](https://github.com/gesellix/Bose-SoundTouch/pull/657) landed, so your
commit was replayed with two conflicts. It keeps your authorship and date, but
the content is no longer byte-identical to what you pushed. Both resolutions
keep each side rather than choosing:

- `setup.go`: #657 added preflight-warning plumbing around the
  `migrateViaTelnet` call, while this branch removed `targetURL` from its
  signature. Now both.
- `handlers_setup.go`: #657 maps a refused preflight to 409, this branch maps
  an invalid telnet URL to 400. Now a `switch` so both map.

Nothing of yours was dropped. If you want to check, the commit was `fc8cee5`.

## What is good

The URL validation rejects control and whitespace characters, shell
metacharacters, non-http(s) schemes, userinfo, queries and fragments before
any connection is opened. The per-device mutation lock does what it claims.
The documentation is unusually honest, and I want to call that out: sequential
rather than atomic, no filesystem backup, process-local serialisation only, no
restoration of DNS, CA, SSH, account or filesystem state. That is the sort of
thing that saves someone later.

## 1. The revert was offered on speakers that were never migrated

`telnetRevertAvailable` compared the live URLs against
`canonicalBoseTelnetURLs()` and reported "revertable" on any difference. That
set is one original variant, not the only one. This repo's own model of a
factory speaker, `pkg/service/testing/fakespeaker`, uses `stats.bose.com` and
`bmxservice.bose.com` where the canonical set has `events.api.bosecm.com` and
`content.api.bose.io`.

Feeding that fixture to the gate returns true, so the web UI offered "Restore
Bose URLs via Telnet" on a pristine speaker. Pressing it rewrites the device's
genuine factory URLs and commits them through `envswitch`.

The speaker keeps working, since both host sets point at the shut-down Bose
cloud. The cost is that a persisted write is offered on a device nobody
migrated, and it destroys the record of what that device's URLs actually were,
which telnet migration takes no backup of.

The gate now compares each field against the values observed on unmigrated
speakers, normalising case and a trailing slash. Only observed values are
listed: other Bose hostnames appear in the DNS interception lists, but those
record hosts a speaker resolves at runtime, which is not the same as the
configured value of these four fields. A wrong entry there would hide the
revert from someone who needs it, so the list errs toward offering a revert
that turns out to be unnecessary.

`TestCheckIsMigratedFromProbe_CanonicalBoseURLsDoNotOfferTelnetRevert` could
not catch this, since it asserts against the same constants the code compared
with.

## 2. An unrecognised response aborted before the readback that would settle it

This is the class your [amendment note](https://github.com/gesellix/Bose-SoundTouch/pull/647#issuecomment-5392778244)
already addressed once, narrower now rather than gone. I probed
`hasTelnetPersistenceConfirmation` directly:

```
observed shape 1                       accepted=true
observed shape 2                       accepted=true
interleaved device log line            accepted=false
trailing slash normalised by firmware  accepted=false
case difference in echo                accepted=false
banner echoed before confirmation      accepted=false
```

Each `false` aborts with "persistence outcome is uncertain, read back and
reconcile all four URL fields", which is work the service can do itself in one
read-only command. `hasTelnetOKResponse` has the same shape sensitivity for
the runtime writes, and `Probe` does a single `Read` whose error is discarded,
so a banner split across TCP segments can abort on command one.

I kept your fail-fast: no further write is sent, so a rejected sequence cannot
spread. What changed is that `getpdo CurrentSystemConfiguration` now runs
before returning, and the error carries what the device actually holds.
Nothing is reported as successful on ambiguous evidence.

My first attempt at this was wrong and your tests caught it: I let the
read-back decide success, which meant a happy read-back could override an
explicit `NOT OK`. Worth mentioning since it shows those tests earning their
keep.

That did mean narrowing two assertions from "nothing is sent after a
rejection" to "nothing that CHANGES the device is sent", since the read-back
is read-only. If you intended the stronger property for a reason I have
missed, say so.

## 3. A rejected URL was reported as telnet being unavailable

Adding `urls.validate()` to `setAllBoseURLsViaTelnet` meant
`resyncBoseURLsAfterXML` turned a validation rejection into "could not re-sync
boseurls over telnet", which reads as an unreachable device. `applyURLOverrides`
lets the XML path accept URLs the telnet validator refuses, so the XML write
succeeds while the re-sync is skipped. A reboot reconciles from the XML, so
the outcome was fine and only the diagnosis was wrong.

---

Also worth flagging, from reviewing #657: your `/full` preset-filtering
observation pointed at something bigger. Capturing sources, presets and
recents before migration would fit at `BackupConfigOffDevice`, which already
runs as step 0, though source credentials are not available over HTTP. Tracked
separately, not for this PR.

PR #649, opened 2026-08-24
fix(player): support browsers without import maps

Mr-Tao:
## Summary

- Replace the player import map and bare module specifiers with direct local ES module URLs.
- Keep the Preact Hooks peer import local when vendored dependencies are refreshed.
- Add a regression test that rejects import maps and bare frontend dependency imports.

This lets browsers such as Safari on iPadOS 15 load the embedded player. They support native ES modules, but not import maps.

## Linked issue

N/A - small compatibility fix based on a blank-page reproduction on iPadOS 15.8.8.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

Validation performed with Go 1.27.0 in an Arch Linux container:

- `go vet ./...`
- `go test ./pkg/service/soundtouchweb ./cmd/soundtouch-player ./cmd/soundtouch-service`
- `go test ./... -skip '^TestHandleTuneInPlayback_Authorized$'`
- static ARM64 builds of `soundtouch-service` and `soundtouch-player`
- `shellcheck scripts/update-static-deps.sh`

The excluded test calls the live TuneIn endpoint and reproducibly received an HTTP 500 response with a non-JSON body. The same failure is unrelated to these static frontend changes. `golangci-lint` 2.12.2 could not complete under Go 1.27 because its `honnef.co/go/tools` analyzer panicked in `buildir`; it did not report a finding in this change.

The ARM64 player built from this commit was deployed through the existing UniFi addon and passed its checksum, version, HTTP, and device-inventory health checks. The embedded player UI then loaded successfully in Safari on an iPad running iPadOS 15.8.8, where the previous build showed a blank page.

## Checklist

- [x] Docs or CLI help updated if behavior changed (not applicable; no configuration or CLI behavior changed)

---

By contributing, you agree that your work is licensed under the project's [MIT License](https://github.com/gesellix/Bose-SoundTouch/blob/main/LICENSE) and that you will follow the [Code of Conduct](https://github.com/gesellix/Bose-SoundTouch/blob/main/CODE_OF_CONDUCT.md).


pkg/service/soundtouchweb/static/js/components/Announcements.js
```diff
@@ -1,6 +1,6 @@
-import { h } from 'preact';
-import { useState, useEffect } from 'preact/hooks';
-import htm from 'htm';
+import { h } from '/app/static/lib/preact.module.js';
```
gesellix: Unsure about the paths - there might be a difference whether to start with a slash `/` or without one.

pkg/service/soundtouchweb/static/index.html
```diff
@@ -5,15 +5,6 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>AfterTouch</title>
         <meta name="description" content="AfterTouch — A replacement for Bose SoundTouch cloud services. Control your speakers after the cloud shutdown." />
-        <script type="importmap">
```
gesellix, line 8: Have you considered this shim? https://github.com/guybedford/es-module-shims

Would you be able to give it a try? I'd prefer to keep compatiblity layers for older devices/browsers more encapsulated.

pkg/service/soundtouchweb/static/js/components/Announcements.js
Mr-Tao: Good point. I changed this in `fce8bb7`: components now import from `../dependencies.js`, while that small facade resolves the vendored modules through relative `../lib/...` URLs. Resolution is therefore relative to the module files and no longer depends on a leading `/` or a deployment prefix. Focused route tests cover both the standalone and embedded mounts, and the exact ARM64 build rendered successfully on iPadOS 15.8.8.

pkg/service/soundtouchweb/static/index.html
Mr-Tao: Yes, I considered `es-module-shims`. Safari 15 already supports standard ES modules; import maps are the unsupported part here. Adding the shim would introduce another runtime loader for only three vendored dependencies.

In `fce8bb7` I instead encapsulated those paths in `dependencies.js`: the app and components import only from that facade, and only the facade knows the vendored file locations. This keeps the compatibility detail localized without adding a runtime shim. The focused Go tests, race test, `go vet`, JavaScript syntax checks, modern Chromium smoke test, and an actual iPadOS 15.8.8 canary all pass.

pkg/service/soundtouchweb/static/lib/es-module-shims.js
```diff
@@ -0,0 +1,1486 @@
...
+    function cleanup() {
+      for (const key of registryKeys.slice(curIdx, (curIdx += 100))) {
+        const load = registry[key];
+        if (load && load.b && load.b !== load.u) URL.revokeObjectURL(load.b);
+      }
+      if (curIdx < registryKeys.length) handler(cleanup);
+    }
+  };
+
+  const urlJsString = url => `'${url.replace(/'/g, "\\'")}'`;
```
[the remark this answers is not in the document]
gesellix: Real bug in the vendored `es-module-shims.js` (2.8.4) itself — `urlJsString` escapes `'` but not a preceding backslash first, which can break out of the generated string literal if a resolved module URL ever contains a backslash immediately followed by `'`. Traced this by hand rather than take CodeQL's word for it — it's a genuine escaping-order bug, not a false positive.

Not reachable in this project though: we never build a dynamic `import()` specifier from untrusted input, which is the precondition for exploiting it. Since we vendor this file verbatim from npm and don't patch third-party code locally (any hand-edit would just get overwritten by `scripts/update-static-deps.sh` on the next refresh), I excluded `pkg/service/soundtouchweb/static/lib/**` from JS/TS CodeQL analysis in `.github/codeql-config-js.yml` instead, mirroring the existing Go config's `paths-ignore` for vendored/generated code.

pkg/service/soundtouchweb/static/lib/es-module-shims.js
```diff
@@ -0,0 +1,1486 @@
...
+      const importMapTest = `<script nonce=${nonce || ''}>${
+      policy ? 't=(window.trustedTypes||window.TrustedTypes).createPolicy("es-module-shims",{createScript:s=>s});' : ''
+    }b=(s,type='text/javascript')=>URL.createObjectURL(new Blob([s],{type}));c=u=>import(u).then(()=>true,()=>false);i=innerText=>document.head.appendChild(Object.assign(document.createElement('script'),{type:'importmap',nonce:"${nonce}",text:${
+      policy ? 't.createScript(innerText)' : 'innerText'
+    }}));i(\`{"imports":{"x":"\${b('')}"}}\`);i(\`{"imports":{"y":"\${b('')}"}}\`);cm=${
+      supportsImportMaps && jsonModulesEnabled ? `c(b(\`import"\${b('{}','text/json')}"with{type:"json"}\`))` : 'false'
+    };sp=${
+      supportsImportMaps && wasmSourcePhaseEnabled ?
+        `c(b(\`import source x from "\${b(new Uint8Array(${JSON.stringify(wasmBytes)}),'application/wasm')\}"\`))`
+      : 'false'
```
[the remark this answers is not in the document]
gesellix: Same situation as the incomplete-sanitization finding on this file: real bug in the vendored library — the CSP-nonce feature-detection helper interpolates `nonce` into an unquoted HTML attribute (`<script nonce=${nonce || ''}>`) without escaping it. Not reachable here though, since this project never sets a CSP nonce or `esmsInitOptions.nonce` anywhere, so `nonce` is always empty. Covered by the same `.github/codeql-config-js.yml` exclusion.

pkg/service/soundtouchweb/static/lib/es-module-shims.js
```diff
@@ -0,0 +1,1486 @@
...
+    };Promise.all([${supportsImportMaps ? 'true' : "c('x')"},${supportsImportMaps ? "c('y')" : false},cm,${
+      supportsImportMaps && cssModulesEnabled ?
+        `cm.then(s=>s?c(b(\`import"\${b('','text/css')\}"with{type:"css"}\`)):false)`
+      : 'false'
+    },sp,${
+      supportsImportMaps && wasmInstancePhaseEnabled ?
+        `${wasmSourcePhaseEnabled ? 'sp.then(s=>s?' : ''}c(b(\`import"\${b(new Uint8Array(${JSON.stringify(wasmBytes)}),'application/wasm')\}"\`))${wasmSourcePhaseEnabled ? ':false)' : ''}`
+      : 'false'
```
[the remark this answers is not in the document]
gesellix: Same root cause as the other two findings on this file — this time `nonce` gets interpolated into a double-quoted JS string literal (`text:"${nonce}"`) without escaping, which could break out of the generated `<script>` body. Not reachable in this project for the same reason as the sibling finding: no CSP nonce is ever set here. Covered by the same `.github/codeql-config-js.yml` exclusion.

gesellix:
Thanks for the contribution — tracking down a real Safari/iPadOS 15 repro and getting a working fix together is genuinely appreciated.

I pushed a few follow-up commits that keep your fix but change its shape a bit: restores the import map and bare-specifier imports everywhere, and lets `es-module-shims` handle the compatibility gap directly (feature-detected, since it's a ~80KB download and most browsers don't need it). Also added two opt-in browser tests (`make test-browser`, chromedp) to exercise the old-Safari path without needing real hardware, plus a separate `browser-tests.yml` workflow.

`go build`/`vet`/`test -race` and `golangci-lint` are clean. I don't have iPadOS 15 hardware myself, so it'd be great if you could re-confirm the original repro is still fixed when you get a chance. Let me know if anything here doesn't match what you had in mind.

Mr-Tao:
Thanks for the thorough follow-up. I tested the merged `734b921` production page on the original iPadOS 15.8.8 device, including the feature-detected dynamic `es-module-shims` path. The player renders correctly, so the original blank-page regression is fixed.

For completeness, the original direct-import approach still had some architectural advantages: it required no runtime loader or vendored minified JavaScript, needed no CodeQL exclusion for `static/lib/**`, and kept the tested path identical to normal production loading. The merged approach preserves the existing import-map module graph and minimizes changes to the application modules, so I understand that trade-off.

One coverage clarification: `TestPlayerRendersUnderForcedShimMode` verifies the vendored shim and module graph under Chrome, but its explicit shim mode does not exercise the shipped dynamic loader or Safari timing. The successful real-device canary now provides that missing target-browser coverage.

Issue #651, opened 2026-08-25
Embedded player does not recover persisted devices after an early startup probe fails


Mr-Tao:
## What happened?

When `soundtouch-service` starts before its persisted SoundTouch devices are reachable, the embedded player's one-shot projection of datastore devices can fail. Making the devices reachable afterwards does not populate the player inventory until an operator triggers discovery or restarts the service.

The service datastore itself remains intact. The stale state is limited to the embedded player's in-memory device registry.

## How can it be reproduced?

1. Keep at least one device with an IP address in the service datastore.
2. Start `soundtouch-service` while that device's `/info` endpoint is temporarily unreachable.
3. Make the endpoint reachable after the initial probe has failed.
4. Read `/api/control/devices/` without triggering discovery.

The persisted device remains absent from the player inventory.

## What did you expect?

The embedded player should retry only the persisted, still-missing device addresses for a bounded startup period, stop as soon as they are registered, and not enable mDNS or UPnP discovery as a side effect.

## Environment and observations

- Current `main`, complete `soundtouch-service` with the embedded player.
- Linux ARM64 gateway with discovery deliberately disabled.
- On a real cold boot, the service became healthy before the LAN devices were routable. Their initial probes failed with `no route to host`.
- A candidate bounded retry registered every persisted device automatically on a later attempt, about 80 seconds after service start, without a discovery request or service restart.

No real LAN addresses, device identifiers, account identifiers, or raw logs are included in this report.

gesellix:
The PR #652 has been merged with some adjustments.

gesellix:
This should be fixed with [v0.130.0](https://github.com/gesellix/Bose-SoundTouch/releases/tag/v0.130.0).

PR #652, opened 2026-08-25
fix(service): retry persisted player devices during startup

Mr-Tao:
## Summary

The embedded player currently projects datastore-backed devices only once at service startup. On systems where `soundtouch-service` starts before the LAN is ready, those probes can fail and leave the player inventory empty or partial until an operator triggers discovery or restarts the service.

This change:

- retries only missing hosts returned by the embedded service's `ExtraDeviceHosts` hook every 30 seconds for at most 10 minutes;
- stops immediately once all currently persisted hosts are registered;
- keeps mDNS and UPnP discovery disabled;
- avoids refreshing `LastSeen` for already registered hosts; and
- protects a newer replacement connection from cleanup by an older in-flight probe.

The standalone `soundtouch-player` does not configure `ExtraDeviceHosts`, so its behavior is unchanged.

## Linked issue

Refs #651

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

Local validation of the exact PR commit:

- `gofmt -s -d` produced no diff;
- `golangci-lint 2.13.1 run` reported zero issues;
- `go test ./pkg/service/soundtouchweb`;
- `go test -race ./pkg/service/soundtouchweb`;
- `go vet ./pkg/service/soundtouchweb ./cmd/soundtouch-service`; and
- `go test ./... -skip '^TestHandleTuneInPlayback_Authorized$'`.

The excluded test requires a live authorized TuneIn account. The Docker-based HTTP-client part of `make check` was not available locally because its required registry is blocked; CI is expected to exercise it.

On a real Linux ARM64 cold boot, every initial persisted-device probe failed with `no route to host`. Without a discovery request or service restart, all persisted devices were registered automatically by a later retry about 80 seconds after service start. The inventory then remained stable.

## Checklist

There is no new user-facing option or CLI surface to document. The bounded startup behavior is documented at its service wiring and public helper.

---

By contributing, I agree that this work is licensed under the project's [MIT License](https://github.com/gesellix/Bose-SoundTouch/blob/main/LICENSE) and that I will follow the [Code of Conduct](https://github.com/gesellix/Bose-SoundTouch/blob/main/CODE_OF_CONDUCT.md).


gesellix:
Thanks for this, the bounded retry approach is a good fix for the cold-boot
race described in #651.

I pushed a follow-up commit on top of your branch (with "allow edits from
maintainers" enabled) that addresses a few things a closer review turned up:

- A datastore read failure in the `ExtraDeviceHosts` hook could make the
  readiness check trivially pass (empty host list looks the same as "all
  hosts registered"), ending the retry window early on a transient glitch.
- Stale-host pruning only looked at hosts inserted during the *current*
  attempt, and only ran inside the bounded retry loop, not the plain
  `SeedExtraDevices()` path used by the devices-changed hook and the manual
  Discover button.
- The retry loop and a devices-changed-hook-triggered reseed could probe the
  same offline host concurrently.
- The unconditional startup broadcast got dropped when every attempt in the
  window inserted nothing.
- The retry interval/window are now `--device-seed-retry-interval` /
  `--device-seed-retry-window` flags instead of hardcoded constants.

Also simplified the channel-based fan-in in `seedExtraDevices()` to a plain
mutex-protected slice.

`go build`/`go vet`/`go test ./... -race` and `golangci-lint` are clean.
Let me know if anything here doesn't match what you had in mind.

Issue #252, opened 2026-05-11
[Feature]: Add support for ST10 stereo pair configuration (as before in Bose App)


BirdyBA:
### Library Version

v0.75.0

### Problem Description

I have 2 ST10 in a room and I want to configure them as stereo pair as before 
(for security reasons I separated the pair before Bose sunset the cloud support)

### Proposed Solution

Stereo pair configuration is saved in:
/mnt/nv/BoseApp-Persistence/1/GroupService.xml

Currently it is automatically reset to an empty file after a reboot.

Is it possible to add / support this file with soundtouch-service like the other files Presets.xml etc.

https://wiki.fhem.de/wiki/BOSE_SoundTouch_de-clouding

### Component

Client Library (pkg/client)

### Device Compatibility

SoundTouch 10

### Priority

Low - Nice to have

### API Type (if applicable)

None

### Use Case / User Story

 Add support for ST10 stereo pair configuration (as before in Bose App)

### Desired API Example

```go

```

### Implementation

- [x] I can help implement this feature
- [x] I can provide testing/feedback
- [ ] I can help with documentation
- [ ] I need someone else to implement this

### Research & References

https://wiki.fhem.de/wiki/BOSE_SoundTouch_de-clouding

### Checklist

- [x] I have searched for existing issues and feature requests
- [x] I have checked the [API Cookbook](docs/API-COOKBOOK.md) for existing functionality
- [x] This feature is related to Bose SoundTouch functionality
- [ ] I have considered backward compatibility

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gesellix:
I'd be happy to add the required cli and service implementation, but I can't test it due to a lack of devices: as far as I know this is specific to adding two ST10 speakers together and I won't be able to group a ST10 and a ST20, is that correct?

gesellix:
With [v0.77.0](https://github.com/gesellix/Bose-SoundTouch/releases/tag/v0.77.0) group management based on the linked api reference has been added to the soundtouch-cli. This is some kind of blindflight, as I already mentioned.

The `group` subcommand should work on your ST10 speakers, I can only test the status on my ST10 while running into timeouts on my ST20. Example:

```shell
# soundtouch-cli --host 192.168.123.123 group status                
Getting group information from 192.168.123.123:8090...
Device is not in a stereo pair
```

[...]

PR #654, opened 2026-08-26
fix(player): project stereo pairs as logical devices

Mr-Tao:
## Summary

soundtouch-player currently lists both physical members of a stereo pair as independent control targets. This change projects a validated two-speaker `LEFT`/`RIGHT` group as one logical target keyed by its master speaker.

The implementation:

- tracks `/getGroup` state alongside the existing device status and applies `groupUpdated` events without allowing an older poll to overwrite a newer event;
- polls the SoundTouch 10-only `/getGroup` endpoint only for models that support stereo pairing, so ordinary refresh remains responsive on SoundTouch 20 and SoundTouch 30 speakers;
- collapses only exact two-member groups whose registered members agree on the group claim, while malformed or conflicting data fails open to the physical entries;
- uses the same logical-device projection for the REST list and global player WebSocket while retaining the established per-device `status_update` frames;
- captures each device's discovery timestamp under the registry lock before projecting it, avoiding a race with concurrent discovery refreshes;
- reports member availability and a degraded state to the player; and
- leaves the raw physical-device registry and pair lifecycle unchanged.

The player intentionally does not offer its existing single-device remove action for a projected pair. Pair creation, rename, and dissolve remain available through the existing client and CLI group operations.

This is the projection prerequisite for #665, #666, and #655. Until it merges, GitHub comparisons for those stacked PRs also include these five commits.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

The original four patches were rebased onto current `main` without semantic changes (`git range-diff` reports all four as identical), followed by the timestamp-snapshot race fix. At the rebased head `332a72f2a743c4d13b7f8bd18df2ff0294fc5e23` and in the integrated stack:

- focused `soundtouchweb` and `webtypes` tests passed;
- the same packages passed under the race detector;
- `go vet ./...` passed;
- `golangci-lint` 2.13.2 passed in exact-head GitHub CI;
- both native and forced-shim browser tests passed three consecutive runs; and
- `git diff --check` passed.

The integrated full `go test ./...` run passed every package except the existing environment-sensitive `TestUnifiedDiscoveryCache`, which received changing real-LAN SSDP/mDNS results between its two 100 ms discovery windows (five devices, then none); the isolated rerun passed.

The original patch-equivalent candidate also had all upstream checks green. Host-gated integration tests self-skipped because `SOUNDTOUCH_TEST_HOST` was unset.

A static Linux ARM64 build of the patch-equivalent candidate was deployed as a canary on a UniFi gateway with a real stereo pair made from two SoundTouch 10 speakers. The player exposed one logical target under the master, reported `Stereo pair 2/2`, retained both projected `devices` and legacy `status_update` WebSocket frames, and rendered correctly in desktop, tablet, and iPhone checks. Both physical speakers remained independently readable and their persisted group data stayed byte-identical.

The model guard was exercised on real SoundTouch 20 and SoundTouch 30 speakers. Four complete refresh rounds over 96 seconds kept `/info`, `/getZone`, `/recents`, and `/now_playing` responsive. A simultaneous 40-second gateway capture contained the normal player refresh requests and no `/getGroup` request, while `/getGroup` remained available for the SoundTouch 10 pair.

## Checklist

This PR does not add pair creation, rename, dissolve, or multiroom-zone behavior. It also does not change the raw device registry or address the separately tracked global WebSocket writer lifecycle.

### DIFF
--- a/cmd/soundtouch-service/testdata/router_routes.txt
+++ b/cmd/soundtouch-service/testdata/router_routes.txt
@@ -46,6 +46,7 @@ GET      /api/control/devices/{id}/power-status                       soundtouch
 GET      /api/control/devices/{id}/recents                            soundtouchweb.(*WebApp).HandleDeviceRecents-fm
 GET      /api/control/devices/{id}/ws                                 soundtouchweb.(*WebApp).HandleDeviceWebSocket-fm
 GET      /api/control/devices/{id}/zone/                              soundtouchweb.(*WebApp).HandleGetZone-fm
+GET      /api/control/devices/{id}/zone/candidates                    soundtouchweb.(*WebApp).HandleGetZoneCandidates-fm
 GET      /api/control/providers/library/servers                       soundtouchweb.(*WebApp).HandleDiscoverLibraryServers-fm
 GET      /api/control/providers/radiobrowser/search                   soundtouchweb.(*WebApp).HandleRadioBrowserSearch-fm
 GET      /api/control/providers/tunein/navigate                       soundtouchweb.(*WebApp).HandleTuneInNavigate-fm
--- a/docs/content/docs/appendix/soundtouch-player-roadmap.md
+++ b/docs/content/docs/appendix/soundtouch-player-roadmap.md
@@ -101,13 +101,13 @@ rename and network/firmware info.
 
 ---
 
-## 4. Render stereo pairs as a single device
+## 4. Render stereo pairs as a single device (shipped)
 
-Today soundtouch-player shows the two halves of a stereo pair (formed via
-`/addGroup` — see issue #252) as independent entries in the device list. The
-Bose app collapsed a paired ST10 set into one "L+R" entry; restoring that
-presentation closes the perception gap BirdyBA flagged at
-<https://github.com/gesellix/Bose-SoundTouch/issues/252#issuecomment-4458140305>.
+soundtouch-player projects a valid two-speaker stereo pair (formed via
+`/addGroup` - see [issue #252](https://github.com/gesellix/Bose-SoundTouch/issues/252))
+as one logical control target. This restores the single-entry presentation
+expected by users while preserving both physical speakers in the service
+registry.
 
 **Device API:**
 - `GET /getGroup` on each speaker — returns the current `<group>` with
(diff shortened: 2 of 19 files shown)


gesellix:
Thanks for this. The projection design (timestamp-safe snapshotting, generation-counter ordering against concurrent events, failing open to physical entries on any doubt) is solid work, and this is a real prerequisite for #665/#666/#655 as you noted.

I pushed six follow-up commits from a closer review:

- `Client.GetGroup`'s doc comment claimed non-ST10 devices reply "harmlessly" with an empty group. Verified against real hardware (a live ST20) that this is wrong: it doesn't reply at all; its own firmware eventually returns an explicit `"AllegroWebserver timeout: /getGroup"` error after a delay well past any reasonable client timeout. Also confirmed `/supportedURLs` lists `/getGroup` on that same ST20 regardless, so a future "check supportedURLs instead of the model name" idea would hit the same wall. Noted that in the docs too.
- `IsConnected` could go true purely from `GetGroup` succeeding (a near-guaranteed reply) even when all five real status fetches failed that round, masking a genuine refresh failure on ST10 hardware.
- `sameGroupClaim`'s order-insensitive role comparison and `webtypes.replaceGroup`'s order-sensitive `reflect.DeepEqual` could disagree on whether a pair "changed" if its roles ever arrive in a different order between a poll and a pushed event. Consolidated into one shared `models.SameGroup`.
- A stereo pair's hidden member had become unreachable as a multiroom-zone target (contradicting the "no zone behavior change" note) since `Zone.js` reads the same projected list as everything else. Added a small `GET .../zone/candidates` endpoint that deliberately bypasses the projection: Zone and Group are separate, unrelated groupings, and the backend's own `HandleZoneAdd`/`HandleZoneRemove` already treat them that way.
- The Library tab could silently jump to an unrelated device if the selected one became a hidden pair member mid-session; it now follows it to the pair's master instead.

Two other things I looked at hard turned out to already be correct, deliberate behavior once I checked them against your existing tests. I mention them mainly so nobody "fixes" them later by mistake: `registeredMembersAgree` treating an unregistered member as non-blocking (matches `TestProjectDeviceEntriesShowsDegradedPairWhenMemberIsMissing`), and `status_update` still flowing for a hidden pair member over the WebSocket (matches `TestPeriodicPlayerMessagesPreserveStatusUpdateStream`).

A few lower-priority, non-bug items (the sequential `GetGroup` call, some duplicated device-lookup logic, the new group mutex living alongside the existing lock-free status mechanism) are deliberately left for after the whole stack lands, tracked separately.

gesellix:
The failing test gets a fix in a few minutes, then this pr is good to be merged. The other PRs in this stack will probably need a rebase, since I prefer rebase merges here.

PR #657, opened 2026-08-26
fix(setup): verify account data before migration

Mr-Tao:
## Summary

Speaker migration currently starts changing device configuration without first proving that the service can reproduce the speaker's account and presets. A missing or lossy datastore snapshot can therefore clear presets on the next account sync. In the shared-account case documented in #614, presets were also wiped after reboot even though the rendered `/full` response was correct.

This change makes the service-backed migration path fail closed before any device mutation. The preflight:

- resolves the exact live account and device and requires their persisted `DeviceInfo.xml`;
- distinguishes a valid explicit empty `Presets.xml` from missing, malformed, or legacy data;
- compares preset slot, name, and location across the persisted snapshot, live `/presets`, and the read-only rendered `/full` response;
- refuses migration while `/full` contains any other device; and
- maps readiness failures to HTTP 409 with an actionable remediation message.

Rendering for the preflight is read-only and does not canonicalize or rewrite legacy preset files. Direct CLI callers without a service datastore retain their existing behavior.

## Linked issue

Refs #614

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

- [ ] `make check` passes (the Docker-based part was not available locally)
- [x] `git diff --check`
- [x] `golangci-lint` 2.13.1
- [x] `go vet ./pkg/service/datastore ./pkg/service/marge ./pkg/service/setup ./pkg/service/handlers`
- [x] `go test ./pkg/service/datastore ./pkg/service/marge ./pkg/service/setup ./pkg/service/handlers -count=1`
- [x] Focused race tests for datastore snapshots, migration readiness, and HTTP mapping
- [x] Complete Go suite in a clean Arch container, excluding only the live authorized-TuneIn test

On a real SoundTouch 30 whose rendered account contained three devices, the migration request returned HTTP 409 before telnet or reboot. Read-only captures of `/info`, `/sources`, `/volume`, `/presets`, `/now_playing`, and `/getZone` were byte-identical before and after the request. The speaker's six presets, volume, standby state, empty zone, and all four original service URLs were unchanged.

## Checklist

---

By contributing, I agree that this work is licensed under the project's [MIT License](https://github.com/gesellix/Bose-SoundTouch/blob/main/LICENSE) and that I will follow the [Code of Conduct](https://github.com/gesellix/Bose-SoundTouch/blob/main/CODE_OF_CONDUCT.md).

### DIFF
--- a/docs/content/docs/guides/MIGRATION-GUIDE.md
+++ b/docs/content/docs/guides/MIGRATION-GUIDE.md
@@ -174,6 +174,18 @@ Once the speaker appears, click **Sync Data**. This connects to the speaker and 
 
 Sync pulls the speaker's local state into AfterTouch's datastore, creating an off-device backup of its configuration. If you ran this before May 6, 2026, your account data from Bose's servers was also captured at that time.
 
+Migration is refused until the service has a valid snapshot for that exact
+account and device and verifies that its rendered account data preserves every
+live preset slot. If the migration page asks for Data Sync, sync the device and
+retry instead of bypassing the check.
+
+Migration is also refused while the rendered account contains another device.
+Some speaker firmware wipes its presets after a reboot-triggered resync of a
+shared account even when `/full` contains the correct data. Move the speaker to
+a dedicated account, run Data Sync for it, and then retry migration. Merely
+removing the other devices is not sufficient because discovery can add them
+again before the speaker fetches `/full` after reboot.
+
 ---
 
 ## Step 5: Migrate
--- a/docs/content/docs/guides/MIGRATION-GUIDE.md
+++ b/docs/content/docs/guides/MIGRATION-GUIDE.md
@@ -179,12 +179,14 @@ account and device and verifies that its rendered account data preserves every
 live preset slot. If the migration page asks for Data Sync, sync the device and
 retry instead of bypassing the check.
 
-Migration is also refused while the rendered account contains another device.
-Some speaker firmware wipes its presets after a reboot-triggered resync of a
-shared account even when `/full` contains the correct data. Move the speaker to
-a dedicated account, run Data Sync for it, and then retry migration. Merely
-removing the other devices is not sufficient because discovery can add them
-again before the speaker fetches `/full` after reboot.
+If the account already contains other devices, migration proceeds but the log
+says so. Some speaker firmware has been reported to wipe its presets after a
+reboot-triggered resync of a shared account even when `/full` contains the
+correct data (see issue #614, where the root cause is still open). One account
+holding every speaker in the household is the normal arrangement, so this is a
+warning rather than a refusal; if you do hit the preset wipe, moving that
+speaker to a dedicated account and running Data Sync for it is the known
+workaround.
 
 ---
 
--- a/pkg/service/datastore/datastore.go
+++ b/pkg/service/datastore/datastore.go
@@ -596,6 +596,22 @@ func (ds *DataStore) GetDeviceInfo(account, device string) (*models.ServiceDevic
 	return ds.getDeviceInfoNoLock(account, device)
 }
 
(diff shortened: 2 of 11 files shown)


gesellix:
> refuses migration while /full contains any other device

This is something I wonder why you want to reject a migration: a typical household might have multiple devices for the same account, so adding another device to an existing one with other devices should be totally fine and officially supported.

gesellix:
Thanks for this. The core idea is right and the mechanism is well built: the
check is a pure early return before any device mutation, so it can only
wrongly refuse and never wrongly proceed, and the reads really are read-only
(`ReadPresetSnapshot` and `AccountFullToXMLReadOnly` avoid `GetPresets`'s
legacy-rewrite side effect). `compareMigrationPresets` is a correct bijection
test. Those parts are unchanged.

I have pushed six commits on top of yours rather than editing it, so your
commit is untouched and the changes are reviewable separately. All of them are
about false refusals, since that is where the risk sits when the gate has no
override.

## The one that reverses a decision of yours

Following the question above, I changed the `accountDeviceCount != 1` refusal
into a warning that names the device count and rides along in the migration
log. I did that before you had a chance to answer, which is worth being
explicit about; if there is a reason we have missed, say so and it can come
back in a better form.

The reasoning:

- One account holding every speaker in the household is the normal Bose
  arrangement and is officially supported, so the gate blocks the common case,
  and `MigrateSpeaker` runs the check unconditionally with no override.
- It also misfires on genuinely single-speaker setups.
  `handleDiscoveredDeviceFallback` writes a second device directory keyed by
  the host address under the same account whenever `/info` momentarily fails,
  and its cleanup only runs when `d.SerialNo` is set, which discovery never
  populates. A stale entry left by a DHCP lease change then blocks migration
  permanently.
- Issue #614 concluded the shared-account preset wipe is empirical rather than
  a proven mechanism, with the root cause still open, and the troubleshooting
  entry there is labelled a workaround.
- The guide's own remediation was not reachable in normal use, by its own
  admission: removing the other devices does not help because discovery
  re-adds them.

The provable checks still refuse. Only this one became advisory.

## The one that would have blocked onboarding

Refusing when the live `/info` has no `margeAccountUUID` makes a factory-reset
speaker impossible to migrate. The admin UI migrates first and pairs
afterwards (the "Pairing runs after the URL flip" comment in the setup page),
and MIGRATION-GUIDE.md step 4 tells the user to Generate an account ID on a
factory-reset device. Both now hit a 409 before pairing can run, and the
suggested remedy cannot help: `SyncDeviceData` files an account-less device
under `"default"`, which never matches an empty live account. An unpaired
speaker has no account data to preserve, so the check is skipped there.

## The smaller ones

- A slot cleared through the Marge API leaves a zero-value entry that
  `savePresetsNoLock` persists as `<preset id="">` and `/full` drops, so a
  perfectly synced datastore was refused. Empty slots are now ignored on both
  sides.
- `mapPresetsToFullResponse` deliberately omits a preset whose source is not
  in the account's configured sources. The comparison then failed with "Run
  Data Sync", which cannot bring a source back, so the user looped. That case
  now gets its own action naming the real remedy.
- The `GetExactDeviceInfo` error was discarded and every failure reported as
  "not persisted", so a malformed file or an I/O error was diagnosed as a
  missing sync.
- `GetMigrationSummary` now reports the readiness verdict, so the pre-flight
  panel shows a refusal before the user commits instead of only failing at
  Apply with a 409.

## Still open

Your `/full`-filtering point led somewhere useful: the same missing-source
condition is why presets disappear from the rendered account, and it suggests
capturing sources, presets and recents before migration rather than only
comparing them. `BackupConfigOffDevice` already runs as step 0 and would be
the hook, though source credentials are not available over HTTP, so that needs
its own design. Tracked separately.

If any of the above misreads what you were guarding against, particularly the
device-count case, please push back.

PR #658, opened 2026-08-26
fix(player): snapshot last-seen timestamps

Mr-Tao:
## Summary

- Copy each device's `LastSeen` timestamp into `DeviceSnapshot` while the registry read lock is held.
- Build the HTTP, initial WebSocket, and broadcast device-list payloads through one shared snapshot helper so they cannot race with discovery touching the same connection.
- Add regression coverage for snapshot isolation and concurrent device-list serialization.

The JSON schema and the meaning of `lastSeen` are unchanged. It still records the latest registry discovery or resynchronization, not the latest successful status poll.

## Linked issue

None. This is a small internal race fix found while reviewing a superseded local control-API experiment.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

- [ ] `make check` passes (the Docker-based HTTP integration suite was not run locally)
- [ ] Tested against a real SoundTouch device (no device behavior or wire protocol changes)
- `go test -race -count=20 ./pkg/service/soundtouchweb`
- `go test ./...`
- `go vet ./...`
- `golangci-lint` 2.13.1: zero issues
- `git diff --check origin/main...HEAD`

## Checklist

- [x] Docs or CLI help updated if behavior changed (not required; the existing JSON contract is preserved)

### DIFF
--- a/pkg/service/soundtouchweb/handler.go
+++ b/pkg/service/soundtouchweb/handler.go
@@ -104,11 +104,12 @@ func (app *WebApp) proxyServiceURL() string {
 	return app.ServiceURL
 }
 
-// DeviceEntry pairs a device id with its connection. Used by
-// DeviceSnapshot so callers can iterate without holding the lock.
+// DeviceEntry pairs a device id with its connection and the LastSeen value
+// captured by DeviceSnapshot under the registry lock.
 type DeviceEntry struct {
-	ID     string
-	Device *webtypes.DeviceConnection
+	ID       string
+	Device   *webtypes.DeviceConnection
+	LastSeen time.Time
 }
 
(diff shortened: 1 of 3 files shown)


Mr-Tao:
Upstream main now contains this synchronization fix as part of #654, specifically commit 342cd47 (fix(player): snapshot device timestamps safely), with projection-aware coverage. Rebasing this branch onto current main produces an empty diff. The race-enabled soundtouchweb tests, soundtouch-player tests, and targeted go vet checks pass on that result, so I am closing this PR as superseded by #654 rather than force-pushing an empty branch.

PR #668, opened 2026-08-30
fix(player): harden device state projection

Mr-Tao:
## Summary

The embedded player currently mixes transport identity with presentation data
and treats transient discovery or event-stream failures as immediate device
loss. Device cards also expose removal as a primary navigation action.

This change hardens the projection boundary without changing the persisted
device registry:

- keeps the control identifier separate from the canonical numeric address
  shown and sorted in the player;
- applies live speaker-name updates without allowing an older poll to overwrite
  a newer event;
- derives `online`, `stale`, and `offline` from HTTP reachability, WebSocket
  transport state, and speaker-reported connection state with bounded
  hysteresis;
- generation-fences status, name, and transport updates so delayed work cannot
  replace newer evidence;
- makes WebSocket replacement and device removal cancel the exact transport
  generation they own;
- moves device removal from the navigation card to device detail and removes
  the registry entry only after a successful identity-fenced `DELETE`, treating
  a synchronous datastore reseed as success without deleting a different
  replacement connection; and
- exposes transient player notices through a polite accessible status region.

This PR builds on the stereo projection merged in #654. Its current head is
based directly on upstream `main`, contains only the three hardening commits,
and omits the older patch-equivalent stereo-projection duplicate. It does not
add logical multiroom-zone controls, group-volume behavior, stereo-pair
lifecycle operations, or event publication for unrelated speaker fields.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

The focused `soundtouchweb`, `webtypes`, and WebSocket tests pass on the exact
published head, including the affected race-detector paths and
synchronous-reseed removal regression. `golangci-lint` reports zero issues;
`go vet`, all nine focused frontend tests, JavaScript syntax, ancestry,
identity, and whitespace checks also pass. GitHub CI passes its browser suite,
tests, lint, cross-platform builds including ARM64, Docker build, CodeQL, and
security jobs; the workflow-conditioned integration job is skipped.

An independent review of transport ownership, registry deletion, generation
ordering, and accessibility found no residual issue after the final fixes.
This is synthetic evidence only; the follow-up has not been deployed or
reverified on physical speakers and browsers.

## Checklist

### DIFF
--- a/pkg/client/websocket.go
+++ b/pkg/client/websocket.go
@@ -236,10 +236,12 @@ func (ws *WebSocketClient) connectWithConfig(config *WebSocketConfig) error {
 	defer ws.connectMu.Unlock()
 
 	ws.mu.RLock()
+
 	if ws.connected {
 		ws.mu.RUnlock()
 		return fmt.Errorf("already connected")
 	}
+
 	ctx := ws.ctx
 	dialContext := ws.dialContext
 	ws.mu.RUnlock()
@@ -288,20 +290,25 @@ func (ws *WebSocketClient) connectWithConfig(config *WebSocketConfig) error {
 	ws.mu.Lock()
 	if err := ctx.Err(); err != nil {
 		ws.mu.Unlock()
+
 		_ = conn.Close()
 
 		return fmt.Errorf("WebSocket client closed during connect: %w", err)
 	}
+
 	if ws.connected {
 		ws.mu.Unlock()
+
 		_ = conn.Close()
 
 		return fmt.Errorf("already connected")
 	}
+
 	connection, transportHandler, transportGeneration := ws.activateConnectionLocked(conn)
 	ws.mu.Unlock()
 
 	notifyTransportState(transportHandler, true, transportGeneration)
+
 	go ws.readLoop(config, connection)
 	go ws.pingLoop(config, connection)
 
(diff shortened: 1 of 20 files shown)


gesellix:
Rebased onto current `main` and ran `/code-review --high`, since this is the
biggest change yet to the device-connection state machine.

Your new health/connectivity mechanism turned out to be a genuinely
different concern from the existing per-field `StatusField` fencing, not a
competing reimplementation — one merges payload per field, the other derives
Online/Stale/Offline with hysteresis. Kept both: `CompleteHTTPPoll` is
called with a nil merge, so it only records connectivity metadata without
re-touching payload fields.

Fixed 3 real bugs: a race I introduced while wiring `OnTransportState`
during the rebase (could let a transient WS blip discard a concurrently
succeeding poll); `StereoPair.js` reporting a successful mutation as
"failed" when the post-success refresh hiccuped; `HandleGetZoneCandidates`
still showing a renamed device's stale name. One reviewer finding
(`Disconnect()`'s context-cancellation) turned out to be deliberate,
already-tested behavior — left alone. Two minor findings deferred (not
fixed, low impact): `resolvedDeviceIPAddress`'s one-shot DNS lookup doesn't
retry if it fails at first discovery, and `DiscoverDevices`' per-device loop
is sequential so the new DNS fallback adds a bit more worst-case latency.

`make check` + `-race` on `pkg/service/soundtouchweb`/`pkg/client` +
`golangci-lint` + the frontend unit tests (including your two new
`.test.mjs` files) all pass.

PR #669, opened 2026-08-30
fix(player): enforce same-origin WebSockets

Mr-Tao:
## Summary

Restore Gorilla WebSocket's default same-origin validation for the web player.

`NewWebApp` previously configured `CheckOrigin` to accept every browser origin. This change removes that permissive override, so Gorilla now:

- accepts same-origin browser handshakes;
- rejects cross-origin browser handshakes; and
- continues to accept non-browser clients that omit the `Origin` header.

The player already opens its WebSocket against `location.host`, so its normal browser and PWA path remains same-origin. A focused handshake test covers all three cases.

## Linked issue

Refs #655, #665, and #666.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

- `go test ./pkg/service/soundtouchweb/...`
- `go test -race ./pkg/service/soundtouchweb/...`
- `go vet ./...`
- `golangci-lint` 2.13.2: `0 issues`
- `git diff --check`

No physical speaker is involved in the HTTP WebSocket origin decision.

## Checklist

- [x] No public CLI or documented workflow changes

### DIFF
--- a/pkg/service/soundtouchweb/handler.go
+++ b/pkg/service/soundtouchweb/handler.go
@@ -147,11 +147,12 @@ type DeviceEntry struct {
 
 // NewWebApp creates a new WebApp instance for SPA mode
 func NewWebApp() *WebApp {
-	// Leave Upgrader.CheckOrigin nil to use Gorilla's same-origin policy while
-	// retaining support for non-browser clients that omit the Origin header.
 	return &WebApp{
 		devices:   make(map[string]*webtypes.DeviceConnection),
 		WSClients: make(map[*websocket.Conn]*sync.Mutex),
+		Upgrader: websocket.Upgrader{
+			CheckOrigin: checkWebSocketOrigin,
+		},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


gesellix:
Thanks for this — restoring gorilla's default same-origin check is the right call. Found one real gap in it plus a small test-duplication thing, fixed both on top of your commit.

## Gorilla's default check breaks behind a reverse proxy on a non-default port

Gorilla compares Origin and Host as raw strings, port included. This repo's documented nginx config forwards a portless Host header (nginx's `$host` never includes the port), which is harmless on the default port but breaks every handshake once the proxy's public port is non-default (e.g. `:8443`, a realistic multi-service-hosting shape) — the browser's Origin keeps the port, the forwarded Host doesn't, so it 403s. Verified against gorilla's vendored source and nginx's own docs.

Fix: `checkWebSocketOrigin`/`sameHostIgnoringPort` — same default policy, but hostname-only. Same-origin and cross-hostname behavior unchanged; only a port mismatch on an otherwise-matching hostname is now tolerated. New tests cover both.

## Smaller thing

Your new test re-implemented scaffolding that already exists as `dialTestWebSocket`. Extracted the shared part into `newTestWebSocketServer`, used by both.

PR #670, opened 2026-08-30
fix(player): reconcile source selection state

Mr-Tao:
## Summary

Source selection in the embedded player currently treats an accepted POST as the final outcome. That can leave the UI optimistic after the speaker rejects the source later, and stale source inventory can remain actionable.

This change:

- posts the canonical `source` and `account` pair exactly once;
- reports pending, confirmed, unverified, and failed outcomes from bounded authoritative readback;
- keeps source inventory and now-playing revisions independently ordered across REST and WebSocket updates;
- expires stale source inventory and disables its actions without inventing a newer canonical device revision; and
- preserves the existing response-level API behavior for legacy consumers.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

- [ ] `make check` passes (the Docker-based HTTP integration target was not run)
- [ ] Tested against a real SoundTouch device
- [x] `node --test pkg/service/soundtouchweb/static/js/api.test.mjs`
- [x] `go test ./pkg/service/soundtouchweb/webtypes ./pkg/service/soundtouchweb`
- [x] `go test -race -count=1 ./pkg/service/soundtouchweb/webtypes ./pkg/service/soundtouchweb`
- [x] `go vet ./pkg/service/soundtouchweb/webtypes ./pkg/service/soundtouchweb`
- [x] `go test -tags browsertest -count=1 ./pkg/service/soundtouchweb`

The complete tagged Chromium suite passed in isolation across its desktop, iPhone, and iPad profiles. No real-device result is claimed by this PR.

## Checklist

- [x] No CLI or documentation update is required for this player-state correction

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ GOTEST=$(GOCMD) test
 GOGET=$(GOCMD) get
 GOMOD=$(GOCMD) mod
 GOFMT=gofmt
+NODE?=node
 
 # Build parameters
 BINARY_NAME=soundtouch-cli
@@ -155,6 +156,7 @@ test-coverage:
 test-browser:
 	@echo "Running browser-level compatibility tests..."
 	$(GOTEST) -tags browsertest -v ./pkg/service/soundtouchweb/...
+	$(NODE) --test pkg/service/soundtouchweb/static/js/api.test.mjs
 
 check: fmt vet test test-http-client
 
--- a/pkg/service/soundtouchweb/browser_compatibility_test.go
+++ b/pkg/service/soundtouchweb/browser_compatibility_test.go
@@ -11,14 +11,68 @@ package soundtouchweb
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"io/fs"
 	"net/http"
 	"net/http/httptest"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/chromedp/chromedp"
+	"github.com/gesellix/bose-soundtouch/pkg/service/soundtouchweb/webtypes"
 	"github.com/go-chi/chi/v5"
 )
+
+func newPlayerFixtureServer(t *testing.T, moduleScript string, configure func(chi.Router)) *httptest.Server {
+	t.Helper()
+
+	r := chi.NewRouter()
+	staticFS, err := fs.Sub(StaticFS, "static")
+	if err != nil {
+		t.Fatalf("open static fixture filesystem: %v", err)
+	}
+	r.Get("/app/static/*", http.StripPrefix("/app/static", http.FileServer(http.FS(staticFS))).ServeHTTP)
+	configure(r)
+	r.Get("/fixture", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "text/html")
+		_, _ = fmt.Fprintf(w, `<!doctype html>
+<html><head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<script type="importmap">{"imports":{"preact":"/app/static/lib/preact.module.js","preact/hooks":"/app/static/lib/preact-hooks.module.js","htm":"/app/static/lib/htm.module.js"}}</script>
+<link rel="stylesheet" href="/app/static/css/app.css">
+</head><body><div class="app"><div id="fixture"></div></div>
+<script type="module">%s</script></body></html>`, moduleScript)
+	})
+
+	server := httptest.NewServer(r)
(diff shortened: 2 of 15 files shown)


gesellix:
Thanks for this. The problem you identified is real and I hit it on my own
hardware while reviewing: a `/select` answers 200 even when the speaker does
not act on it, and the player had no way to tell. I have taken this forward
as [PR #687](https://github.com/gesellix/Bose-SoundTouch/pull/687), rebased
onto `main` with follow-up commits.

Your commit is preserved with your authorship as the first commit, and
everything I changed is a separate commit on top, so the diff between your
work and mine is visible in the log. I would rather be explicit about what I
altered, because several of those changes rest on assumptions you are better
placed to check than I am.

## The one change inside your commit

The rebase could not take your work unmodified. Your change introduces its
own per-field fencing (`fieldRevision`, `deviceStatusFieldRevisions`, the
`MergeNowPlaying` / `MergeVolume` / `MergePresets` / `MergeSources` /
`MergeBass` / `MergeIsConnected` methods) over the same six fields that
`main`'s `StatusField` mechanism (`BeginFieldPoll` / `CompleteFieldPoll` /
`ApplyFieldEvent`) already orders. That mechanism landed after you branched.

Landing both would have left two independent generation counters guarding
the same state, which is a bug class that stack already had to fix three
times. I resolved it in favour of the existing mechanism, so the `Merge*`
methods are gone and the WebSocket handlers use `main`'s calls. Your public
`Revision` and `NowPlayingRevision` are kept, with `NowPlayingRevision`
derived from the existing `FieldNowPlaying` generation rather than a second
counter. The reasoning is in that commit's message.

Everything else in your commit is unmodified.

## What I changed afterwards, and why

- **Sources staleness.** The 30s TTL evaluated inside `Status()` matched the
  30s poll ticker exactly, so the buttons would have gone disabled shortly
  before every refresh. It now marks stale after two consecutive failed
  `/sources` reads instead, which also removed the need for
  `mergeDerivedStatus`.
- **Write failure classification.** `checkedReq` treated every failure alike.
  I split it into definitive (4xx, raised before the speaker is contacted)
  and ambiguous (5xx or transport, where a reply lost after the speaker
  already switched is indistinguishable from one it never received). Worth
  saying that I got this backwards on the first attempt and a review caught
  it: `sendControlResponse` maps a failed `SelectSource` to 500, so my first
  version failed fast on exactly the ambiguous case.
- **Readback cost.** A confirming readback did not stop the later ones, and
  each fetched the whole device (a full status poll, six speaker calls). A
  confirmed selection cost around 21 speaker requests. The later readbacks
  are now kept only when the readback reports no live event stream, and they
  use a new `/devices/{id}/now-playing` endpoint that reads one thing.
- **Smaller ones.** Restored the `ready.length === 0` early return, so a
  device that has not been polled no longer renders an empty section;
  `api.test.mjs` moved to `frontend_test/`, where `make test-frontend` and
  the CI job already pick it up, so the `Makefile` change was reverted.

## Where your assumptions and mine might differ

These are the parts I would most like you to check against your setup.

1. **The early stop depends on `webSocketConnected`.** On one of my runs it
   was `false` for every speaker while connectivity was `online`, and I never
   found out why. In that state the optimisation does nothing and every click
   takes the full three readbacks. If your speakers sit in that state, you
   will see different numbers than I describe.
2. **`TUNEIN` is inferred, not tested.** Hardware testing turned up a
   separate problem: `RADIO_BROWSER` and `LOCAL_INTERNET_RADIO` are
   advertised `READY` but are not selectable inputs, and a bare select parks
   the speaker on a stub now-playing while the previous audio keeps playing.
   I verified both directly. I treat `TUNEIN` the same way because
   `stations.ResolveContentItem` handles it identically, but I have not
   confirmed it on a speaker. If TuneIn behaves differently for you, that
   assumption is wrong.
3. **`ALEXA` is left alone** for the same reason, unverified. It is covered
   only by a generic check that refuses to confirm a stub now-playing.
4. **Two consecutive failures before marking sources stale** is a judgement
   call, chosen to match how connectivity is already debounced. If you have
   seen `/sources` fail differently, that number may want changing.
5. **Provider sources now resume from Recents** rather than issuing a bare
   select, except `LOCAL_INTERNET_RADIO`, where AfterTouch's own TTS and
   notification ding go through the same source and resuming played the ding.

If any of that contradicts what you see, please say so. Notes are in
`docs/content/docs/reference/PLAYER-SOURCE-BEHAVIOUR.md` in the new PR.

Thanks in general for all the work on the player through the several pull requests!

gesellix:
Merged with https://github.com/gesellix/Bose-SoundTouch/pull/687.

PR #673, opened 2026-08-31
feat(client): add capability-driven device settings

Mr-Tao:
## Summary

The SoundTouch HTTP client already exposes playback and grouping controls, but
device settings are still represented by ad-hoc requests in higher layers.
That makes capability-driven settings difficult to validate and reuse outside
one UI.

This change adds the typed client and model foundation for device settings:

- models and validates automatic standby, rebroadcast latency, language,
  Bluetooth information, clock display, network information, and source-name
  payloads;
- exposes explicit client read and mutation methods for the corresponding
  speaker endpoints, including dedicated Bluetooth pairing and clear-list
  actions;
- sends firmware-required state-changing GETs without transport replay,
  validates their endpoint-specific status response, and distinguishes an
  authoritative speaker rejection from an unknown mutation outcome;
- rejects incomplete or unsupported XML values instead of silently accepting
  ambiguous state; and
- preserves whether the firmware actually reported the clock enable state, so
  higher layers cannot turn an omitted field into a writable false value; and
- makes setup and normal client use share the same validated SoundTouch
  language-code model.

The PR is exact commit `993bbe8`, directly based on current upstream `main`.
It does not add embedded-player routes or UI, does not invoke any mutation on
its own, and does not include the separately prepared acoustic or network
diagnostic features.

## Linked issue

Refs #252

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

On exact commit `993bbe8`:

- focused model, client, setup, and CLI tests pass;
- the affected model and mutation paths pass under the race detector;
- `golangci-lint` v2.13.2, `go vet`, a CLI build, formatting checks, and
  `git diff --check` pass for the affected scope; and
- focused regressions cover typed HTTP-200 error envelopes, redirect
  non-replay, lost and unstructured response classification, missing language
  values, clock-enable presence and omission, defensive language-map copies,
  and the English language code used by setup.

The endpoint shapes and alternate/restored values were previously verified on
real SoundTouch 10, 20, and 30 speakers through the integrated canary. This
focused foundation commit has no player surface and has not itself been
deployed as a standalone build.

## Checklist

### DIFF
--- a/cmd/soundtouch-cli/cmd_setup.go
+++ b/cmd/soundtouch-cli/cmd_setup.go
@@ -1979,7 +1979,7 @@ func setupPairCmd() *cli.Command {
 			&cli.StringFlag{Name: "mode", Value: "full", Usage: "full (state machine) or bare (setMargeAccount only — experimental)"},
 			&cli.StringFlag{Name: "service-url", Value: "http://aftertouch.local:8000", Usage: "AfterTouch base URL (also populates <boseServer>/<updateServer> in setMargeAccount)"},
 			&cli.StringFlag{Name: "name", Usage: "Speaker name to set during pairing (empty = keep current)"},
-			&cli.IntFlag{Name: "language", Value: setup.LanguageEnglish, Usage: "sysLanguage code (2 = English)"},
+			&cli.IntFlag{Name: "language", Value: setup.LanguageEnglish, Usage: "sysLanguage code (3 = English)"},
 			&cli.DurationFlag{Name: "step-timeout", Value: 8 * time.Second},
 			&cli.StringFlag{Name: "token", Usage: "userAuthToken value (empty = use built-in placeholder matching the Bose app token shape)"},
 		},
--- a/docs/content/docs/analysis/SETUP-WEBSOCKET-EXPERIMENT.md
+++ b/docs/content/docs/analysis/SETUP-WEBSOCKET-EXPERIMENT.md
@@ -120,6 +120,10 @@ This drives `setup.Manager.ExecuteInitPlan` with `SkipURLRewrite=true`, which ru
 > `SOUNDTOUCH_NOT_CONFIGURED`, so the preflight passes through unchanged;
 > see `docs/content/docs/reference/DEVICE-PAIRING-FLOW.md`.
 
+The historical capture below sent language code `2`. Stockholm's language
+table identifies that code as German; current English-language setup uses code
+`3`. The original wire value is retained here as experiment evidence.
+
 ```
 SETUP_START
 SETUP_IDENTIFY_DEVICE_ENTER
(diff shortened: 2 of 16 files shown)


Mr-Tao:
The browser job failed while starting Chromium (`websocket url timeout reached`) before the native render assertion ran. The forced-shim render and all subsequent tests in the same job passed; the exact head also passes `make test-browser` locally, and the corresponding workflow is green on `main`.

GitHub does not allow me to rerun the job as an external contributor. Could you please rerun the failed browser job when convenient?

gesellix:
I'm going to merge https://github.com/gesellix/Bose-SoundTouch/pull/682 first, can you rebase afterwards?

I'll review your pull requests during the weekend!

Mr-Tao:
Thanks! Unfortunately, I won’t be able to get back to this until next week.

gesellix:
Thanks for this — the capability-driven device settings groundwork (system timeout, rebroadcast latency mode, system language, Bluetooth info) is solid, well-tested infrastructure. Rebased onto current `main` and found one real conflict plus a small cleanup on top.

## `mutatingGet`'s signature diverged from a concurrent change on `main`

Between this branch's start and now, `main` changed `mutatingGet` to `(endpoint string, result interface{}) error`, delegating to a shared `getWithHTTPClient` so it can unmarshal a response body (used by `/removeGroup`). Your version changed it to `(endpoint string) error` with its own inline handling that validates the device echoes back `<status>{endpoint}</status>` — needed for your four new endpoints (`/enterPairingMode`, `/enterBluetoothPairing`, `/clearPairedList`, `/clearBluetoothPaired`), which return no meaningful body to unmarshal.

Both are genuinely needed for different endpoint shapes, so kept both: `mutatingGet` stays as `main` has it, and added `mutatingGetConfirmStatus` for your status-echo case, with your four new call sites pointed at it.

## Also fixed

Splitting that one function into two left ~30 lines of one-shot-transport setup (clone, disable keep-alives, `CheckRedirect`) duplicated between them — my own conflict-resolution artifact, not something in your original commit. Extracted `newOneShotHTTPClient`, shared by both.

## Worth a look when convenient, not a blocker

`EnterPairingMode`/`ClearPairedList` assume the same `<status>` echo pattern as `EnterBluetoothPairing`/`ClearBluetoothPaired`, but only the Bluetooth two have a captured real-firmware example in `UNIMPLEMENTED-ENDPOINTS.md`. It fails closed either way (`ErrMutationOutcomeUnknown`, never a false success), so not blocking — just flagging in case you've verified the legacy pairing/clear-list endpoints against real hardware and can confirm the echo shape.

## Looked at, no action

`systemsettings.go`'s three `UnmarshalXML` implementations look like a repeated skeleton next to the lighter `type TempX X` idiom used elsewhere (`balance.go`, `bass.go`) — I mention it mainly so nobody "simplifies" it later by mistake. They're not interchangeable: your pointer-typed anonymous "wire" structs deliberately distinguish "field absent" from "field present as zero value" to reject incomplete firmware responses, which the lighter idiom can't express. Left as-is.

PR #677, opened 2026-09-03
fix(tunein): profile navigate 500, empty content, and broken episode playback

yongxin-ms:
## Summary

Clicking into a "Show" from TuneIn search results (a Program/Profile item, e.g. searching "bbc" and opening a program) was broken end-to-end. Three separate bugs, found and fixed one at a time against a live SoundTouch speaker:

1. 500 on every profile link. navigateTuneIn / parseTuneInNavigatePath required a profiles/{type}/{id}/{encodedURI} path shape that nothing actually generates (real hrefs are a single profiles/{encodedURI} segment - see tuneInSearchProfile). Every profile link fell through to bmx.TuneInNavigate with the literal "profiles/..." prefix still attached to the string it tried to base64-decode, which reliably failed as invalid base64 and surfaced as a 500.

2. Content list came back empty after the 500 was fixed. TuneInNavigateProfile read a top-level data["pivots"] array with "text"/"URL" fields. The real API nests this under Item.Pivots, keyed by pivot name ({"Item": {"Pivots": {"Contents": {"DisplayName": "Broadcasts", "Url": "..."}}}}) - an object, not an array - so the pivot lookup never matched and bmx_sections came back null. Once that's fixed, the Contents pivot's own response also doesn't list playable items directly: each entry is a "Container" (e.g. a tile titled "Episodes") whose real payload is its Children array, which the old code didn't walk.

3. List displayed, but nothing would play. Individual episode ("Topic") items were linked to /v1/playback/episodes/{guideId} (type tracklisturl). That route is backed by TuneInPodcastInfo, which - per its own existing implementation - ignores the id entirely and always returns an empty track list, so a speaker given that link has nothing to play (confirmed against a real device: POST /play succeeded, but the speaker span its loading indicator forever with no audio). Switched episodes to /v1/playback/station/{guideId} (type stationurl) instead - the one path in this codebase already proven to resolve a raw TuneIn GuideId to a real stream via Tune.ashx (resolveTuneInProgramLatestEpisode + TuneInPlayback already do exactly this for a program's latest topic).

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

- [X] Tested against a real SoundTouch device (details below)
I've tested on my Bose SoundTouch30 II

make check note

make check currently fails on some machines/networks via a pre-existing, unrelated test: TestCheck443Reachability_LANProbeMatchesListenerOutcome in pkg/service/handlers/preflight_test.go (untouched by this PR - last modified in an unrelated earlier commit). It asserts that probing 1.2.3.4:8443 is unreachable, using that address as a "nothing answers here" placeholder; on networks where something along the path intercepts/ responds to arbitrary outbound :443 connections (transparent proxies, DPI, some ISPs), the probe comes back reachable and the assertion fails. Reproduces identically on a clean main with none of this PR's changes, so it's a flaky environmental assumption in that test, not a regression here. The rest of the suite (including this PR's affected packages - pkg/service/bmx, pkg/service/stations, pkg/service/handlers) passes:

`
go test -v ./... -skip TestCheck443Reachability_LANProbeMatchesListenerOutcome
`

## Checklist

- [X] My changes are focused, and I have read the diff myself
- [X] No personal data (real LAN IPs, MAC addresses, device IDs, account IDs) in code, tests, or fixtures

---

### A note on AI-assisted contributions

AI and agent-assisted code is welcome, we use it here too. What we cannot accept is
unreviewed "slop": large generated diffs the author has not read, run, or understood.
Keep PRs small and focused, make sure `make check` passes, and be ready to explain your
changes during review.

By contributing, you agree that your work is licensed under the project's
[MIT License](https://github.com/gesellix/Bose-SoundTouch/blob/main/LICENSE) and that you
will follow the
[Code of Conduct](https://github.com/gesellix/Bose-SoundTouch/blob/main/CODE_OF_CONDUCT.md).

### DIFF
--- a/pkg/service/bmx/tunein.go
+++ b/pkg/service/bmx/tunein.go
@@ -605,20 +605,34 @@ func TuneInNavigateProfile(encodedURI string) (*models.BmxNavResponse, error) {
 		Layout: "classic",
 	}
 
-	// Profiles contain "pivots" (sections like "Programs", "Related", etc.)
-	pivots, _ := data["pivots"].([]interface{})
-	for _, p := range pivots {
+	// The profiles API (api.radiotime.com/profiles/...) nests everything under
+	// "Item", and its pivots ("Contents", "Related", etc.) are an *object*
+	// keyed by pivot name, e.g.:
+	//   {"Item": {..., "Pivots": {"Contents": {"DisplayName": "Broadcasts", "Url": "..."}}}}
+	// (Earlier code assumed a top-level "pivots" array with "text"/"URL"
+	// fields, which never matched this response and left bmx_sections empty.)
+	item, _ := data["Item"].(map[string]interface{})
+	pivots, _ := item["Pivots"].(map[string]interface{})
+
+	for pivotName, p := range pivots {
+		// We only care about the "Contents" pivot for now (the main list).
+		if !strings.EqualFold(pivotName, "contents") {
+			continue
+		}
+
 		pivot, ok := p.(map[string]interface{})
 		if !ok {
 			continue
 		}
 
-		pivotName, _ := pivot["text"].(string)
-		pivotURL, _ := pivot["URL"].(string)
-
-		// We only care about the "Contents" pivot for now (the main list)
-		if !strings.EqualFold(pivotName, "contents") {
+		pivotURL, _ := pivot["Url"].(string)
+		if pivotURL == "" {
 			continue
+		}
+
+		displayName, _ := pivot["DisplayName"].(string)
+		if displayName == "" {
+			displayName = pivotName
 		}
 
 		contents, err := fetchJSON(tuneInRenderJSONURI(pivotURL))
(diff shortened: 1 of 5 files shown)


yongxin-ms:
Hi @gesellix, this PR works fine on my local env and SoundTouch30 II, though there are some failed checks not caused by this PR.

Are you working on these failed checks? If not, could you please merge this PR first, and I'll try to fix them.

Thanks.

gesellix:
@yongxin-ms thanks for the heads-up, I just merged a fix for the security issues. Please rebase.
I'm going to review your pull request during the weekend!

yongxin-ms:
@gesellix Done

> @yongxin-ms thanks for the heads-up, I just merged a fix for the security issues. Please rebase. I'm going to review your pull request during the weekend!

gesellix:
Thanks for this — the two root causes you tracked down (the `pivots` shape assumption, and the `profiles/{type}/{id}/{encodedURI}` path assumption) both check out against the real API responses, and the episode-playback fix is a real improvement. Found a few regressions in the new container-handling logic on top, now fixed.

## The new container logic dropped a few things the code it replaced already handled

`TuneInNavigateProfile` used to call the existing `tuneInSearchSection` helper for each top-level entry; this PR replaced that with new hand-rolled logic, which silently lost:

- **The `Pivots.More.Url` "load more" pagination cursor.** Any profile container with more children than fit on one page (the case your own comment describes) now shows only the first page. Pulled the cursor-link logic already in `tuneInSearchSection` into a shared `tuneInMoreCursorLink` helper, used by both now.
- **The legacy lowercase `"children"` key.** `tuneInSearchSection` already falls back to it when `"Children"` is absent; the new code only checked the capitalized key.

## A container with zero children could produce a bogus playback link

An empty container (no episodes published yet, say) has no `"Type"` field — only leaf items carry that; containers are identified by `"ContainerType"` instead (confirmed via the `tuneInProfileContentsSection` struct already in this file for the same endpoint). The childless-fallback branch checked `"Type"`, got an empty string, and fell into the default case that treats any unrecognized type as directly playable — turning the container's own non-playable `GuideId` into a broken playback link. Empty containers are now skipped instead.

## Two smaller things

- The response's own self link used `/v1/navigate/profile/` (singular) instead of `/v1/navigate/profiles/` (plural) — no route dispatcher recognizes the singular form, so following the response's own `_links.self.href` hit the exact 500 this PR fixes, just via the self link.
- `base64.URLEncoding` (padded) vs. `base64.RawURLEncoding` (no padding), two lines apart, building the same kind of href. `decodeBase64URI` already tries every variant as a fallback (that's why this never surfaced as a decode failure) — standardized both on `RawURLEncoding`, matching every other encode site in the file.

Added `TestTuneInNavigateProfileHandlesContainerShapes` covering all four against a fixture server modeling the real two-fetch profile/contents shape, plus a small fallback fix (in both `handlers_bmx_tunein.go` and `stations.go` — same pre-existing duplicate parsing logic, not something this PR introduced) so an empty encoded URI in the `"profiles"` path falls back like the sibling `"sub"` case already does, instead of a confusing downstream error.

## Looked at, no action

- `parseTuneInNavigatePath`/`navigateTuneIn` are now byte-for-byte identical between `handlers_bmx_tunein.go` and `stations.go` — confirmed both already independently had a (buggy) `"profiles"` case before this PR, so you fixed the same bug consistently in both rather than introducing new duplication. Worth consolidating (route `HandleTuneInNavigate` through `stations.Navigate`) at some point, but that's a bigger, separate change than this bug-fix PR should carry.
- The `Station`/`PlayItem`/`Topic`/`Program`/`Profile` type-dispatch switch is now a third/fourth implementation (`tuneInSearchSection`, `TuneInSearchNext`, your new `tuneInProfileNavItem`). Your new default-is-playable fallback is a real, deliberate improvement specific to what your hardware testing found — not something I'd want to silently apply to the older two without evidence, so left as three separate switches for now.

PR #685, opened 2026-09-05
fix(stereo): stop treating an unreachable external Marge as a hard block

gesellix:
## Summary

Reported against real hardware: creating a stereo pair failed with a 502 and `HTTP 401: Unauthorized` from `streaming.bose.com`, even though both speakers were DNS-migrated to this AfterTouch instance. Root cause was two separate issues in the stereo-pair generation-persistence layer added by the #654/#655 stack:

- **DNS migration isn't detectable from a speaker's reported `MargeURL` alone.** DNS-level migration only changes how a Bose hostname resolves on the network, never the device's own advertised URL, so a DNS-migrated speaker still reports `MargeURL: https://streaming.bose.com`. The generation preflight's `isLocal()` check only matched against this service's own advertised server URL, so it misclassified such a speaker as "external" and queried the *real* Bose cloud instead of the local datastore.
- **The external-Marge path duplicated work the speaker already does itself, unreliably.** A speaker's own firmware self-reports its group create/rename/teardown to whatever Marge backend it's configured with — that's the entire reason `HandleMargeAddGroup`/`HandleMargeModifyGroup`/`HandleMargeDeleteGroup` exist; they're only ever called by speakers, never by us. The coordinator's `GenerationCleanup`/`GenerationRename` proactively pushed the same update to an external backend it generally can't authenticate to (real Bose cloud, another instance), and `GenerationPreflight`'s read-only dangling-generation check hard-blocked `Create`/`Dissolve` if that push/check failed — even though the coordinator's own physical preflight (capability/zone/reachability against the live speakers) is the real safety net.

## Changes

1. Recognize a DNS-hijacked speaker's `MargeURL` as local (via the existing `discovery.InterceptedBoseHosts` list + this service's own DNS-hijack setting), routing it through the reliable local datastore check instead.
2. Stop pushing cleanup/rename to an external Marge backend — no-op, since the speaker self-reports on its own.
3. Make the external preflight's dangling-generation check advisory: still attempted, but a failure (network error, wrong credentials, unreachable backend) no longer blocks `Create`/`Dissolve`.
4. Applied consistently across `soundtouch-service`, the standalone/embedded player, and the CLI.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean
- [x] `go test ./...` full repo pass
- [x] `make lint` — 0 issues
- [x] New unit tests per binary proving: cleanup/rename never call the external backend, preflight still attempts the read but tolerates failure, and a DNS-hijacked Bose hostname resolves to the local (not external) path
- [x] Confirmed against real hardware (Create + Dissolve, DNS-migrated speakers)

### DIFF
--- a/cmd/soundtouch-cli/cmd_group.go
+++ b/cmd/soundtouch-cli/cmd_group.go
@@ -182,19 +182,49 @@ func newGroupCoordinator(config *ClientConfig) *stereopair.Coordinator {
 	}
 
 	cleanupClient := &http.Client{Timeout: lifecycleConfig.Timeout}
+	cleanup, preflight, rename := cliStereoPairGenerationPersistence(cleanupClient)
 
 	return stereopair.NewWithGenerationLifecyclePersistence(
 		groupClientFactory(&lifecycleConfig),
-		func(ref stereopair.GenerationRef) error {
-			return stereopair.DeleteMargeGroupGeneration(cleanupClient, ref)
-		},
-		func(refs []stereopair.GenerationRef) error {
-			return stereopair.EnsureMargeNoGroupGenerations(cleanupClient, refs)
-		},
-		func(ref stereopair.GenerationRef, name string) error {
-			return stereopair.RenameMargeGroupGeneration(cleanupClient, ref, name)
-		},
+		cleanup, preflight, rename,
 	)
+}
+
+// cliStereoPairGenerationPersistence wires generation-lifecycle hooks for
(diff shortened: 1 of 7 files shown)


gesellix:
@Mr-Tao FYI: I'm going to make some changes, which I think won't break anything you provided with your pull requests. Please check and report back if any regression becomes visible in your setup. Thanks!

Issue #646, opened 2026-08-23
[Bug]: On-device install: speaker stuck in source="SETUP" / MargeHSM NotAssociated after migration (SoundTouch 621, spotty)


Henrik509:
## Summary

On-device AfterTouch install on a **SoundTouch 621**. Migration completed 
successfully and radio sources show as READY, but the speaker is stuck in 
`source: "SETUP"` and won't leave onboarding — the display cycles languages 
showing "Download the SoundTouch app", and pressing a preset button produces 
no audio (INVALID_SOURCE).

## Environment

- Device: SoundTouch 621 (reports as type "SoundTouch 20", variant `spotty`)
- Firmware: 27.0.6.46330.5043500
- Install: on-device (aftertouch-service v0.129.0 running on the speaker)
- Device ID / MAC: 000C8A7B107B
- Account: 2814631

## What works

- aftertouch-service running (responds on localhost:8000, LAN via port 17008)
- Migration applied: all four URLs point to http://localhost:8000 
  (marge/stats/swUpdate/bmxRegistry) — Current Config == Planned Config
- Migration panel shows "Account paired 2814631"
- `/sources` shows LOCAL_INTERNET_RADIO, TUNEIN, RADIO_BROWSER as status="READY"
- SystemConfigurationDB.xml has AccountUUID=2814631 and a non-empty 
  AccountAssociatedEMail

## The problem

Despite the above, the speaker never leaves setup. logread shows, repeating 
every ~30s:

    HandleMessage(EVT_CURRENT_SOURCE) >> Setup
    NowPlay: source: "SETUP" ContentItem { source: "SETUP" isPresetable: true }

and earlier:

    MargeStateTop::HandleSyncWithMargeRequest not serviced - MargeHSM is in the wrong state.
    Marge: HandleSyncWithMargeRequest() >> NotAssociated > MargeStateTop
    AVSSTClient: Account ID =        (empty)
    AVSSTClient: No device id. Retrying

## What I tried

- Factory reset + re-migration (to get the radio source types back — that part 
  worked, they're now READY)
- `sourcesUpdated` notification (accepted, no effect on the SETUP state)
- Multiple reboots
- `envswitch accountid set 2814631` via telnet :17000 — accepted ("Setting 
  AccountId to 2814631 / OK"), survives reboot, but speaker still stays in SETUP
- Manually set AccountAssociatedEMail in SystemConfigurationDB.xml (was empty) 
  — persists, but no change

[...]

PR #690, opened 2026-09-07
Make saving to a preset discoverable, and show real station artwork

gesellix:
## Saving to a preset was invisible

The save button on a preset tile was gated twice: it was only rendered while
something was playing, and even then only appeared on hover. On a tablet or
phone, which has no hover, it could not be reached at all.

This came out of the closing comments on
[issue 646](https://github.com/gesellix/Bose-SoundTouch/issues/646), where two
people spent weeks debugging what looked like a speaker fault before
discovering that the thing they wanted was a button they had never seen.

Now the button is always there and simply disabled when there is nothing to
save, and it uses the same gold star as the now-playing card so both ways of
saving read as one gesture: an outline star means "save here", a solid one
means "this is what is playing right now". An empty slot invites the save
directly instead of just reading "Empty", and clicking anywhere on it saves.

## Presets show real station artwork

Radio stations were showing a placeholder even though their logo was available
the whole time. The speaker keeps whatever artwork it is given when playback
starts, and presets, recents and the now-playing card all read it back from
there, but two of the three station play paths were dropping it on the way to
the speaker.

Stations played from now on carry their logo into presets and recents.
Presets saved before this keep their blank artwork until they are saved again.

The now-playing card was also missing artwork for radio stations, because it
only looked at one of the two places a picture can live.

## Preset tiles line up

Tiles were sized by their own content, so a row mixing one-line and two-line
station names came out with ragged bottom edges. They now fill their grid cell,
and the grid itself is responsive: six across on a wide window, three by two in
the device panel, two by three on a phone.

## One icon vocabulary, and readable labels

Presets, Recents and Sources each had their own idea of how to depict a source,
and the two emoji maps disagreed with each other (Amazon was a shopping trolley
in one and a musical note in the other). There is now a single set: sources we
can draw get a monochrome icon that follows the light or dark theme, and the
streaming services keep an emoji rather than shipping their trademarks.

Each source also has a colour, shown along the bottom edge of a preset tile and
its pill in the Sources row, so a source looks the same wherever it appears.

Two contrast problems went with it. Source labels were rendering in near-black
whatever the theme, because buttons do not inherit the page text colour. And
the preset artwork box had a tinted fill behind a mid-tone icon, which left
both hard to make out.

## Transport buttons

The player, pause, skip and toggle buttons were a mix of emoji and drawn icons,
sized by their content. Emoji widths differ per glyph, so no two buttons were
the same size, and the play button visibly changed width every time you paused,
nudging the rest of the row sideways. They are drawn icons in fixed squares
now, and each one finally has a name for screen readers.

## Startup log

`AfterTouch Web UI starting on http://:8080` is not something a terminal will
turn into a link. When the player listens on all interfaces the line now reads
`http://localhost:8080`, while an explicit `--bind` address is left alone.

---

Covered by new tests: the artwork reaching the speaker across all three station
play paths, and guards against the save button regressing back into a
hover-only control.

### DIFF
--- a/cmd/soundtouch-player/browsable_url_test.go
+++ b/cmd/soundtouch-player/browsable_url_test.go
@@ -1,0 +1,31 @@
+package main
+
+import "testing"
+
+// TestBrowsableURL covers the startup log line: a wildcard listen address is
+// what net.Listen wants but is not something a terminal can turn into a
+// clickable link, so it has to be rewritten before it is logged.
+func TestBrowsableURL(t *testing.T) {
+	tests := []struct {
+		name string
+		addr string
+		want string
+	}{
+		{"wildcard", ":8080", "http://localhost:8080"},
+		{"wildcard IPv4", "0.0.0.0:8080", "http://localhost:8080"},
+		{"wildcard IPv6", "[::]:8080", "http://localhost:8080"},
+		{"explicit host", "192.0.2.5:8080", "http://192.0.2.5:8080"},
+		{"loopback", "127.0.0.1:8000", "http://127.0.0.1:8000"},
+		{"hostname", "speaker.local:8080", "http://speaker.local:8080"},
+		{"IPv6 literal stays bracketed", "[2001:db8::1]:8080", "http://[2001:db8::1]:8080"},
+		{"not host:port is left alone", "nonsense", "http://nonsense"},
+	}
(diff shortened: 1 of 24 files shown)


gesellix:
Example screenshot:
image

PR #707, opened 2026-09-10
fix(player): reconcile discrete command outcomes

Mr-Tao:
## Summary

The embedded player currently treats many discrete commands as complete when
their HTTP request returns. A delayed speaker response can therefore leave the
UI optimistic, while retrying a non-idempotent action would risk applying it a
second time.

This change extends the reconciliation model from source selection in `#670`
to the remaining discrete player actions:

- sends power, transport, mute, playback-mode, track, preset, recent, provider,
  URL, and library mutations exactly once;
- reports pending, confirmed, unverified, and failed outcomes from bounded
  authoritative now-playing or status readback;
- starts absolute readback delays without waiting for a slow mutation response;
- fences delayed results when the selected device, command generation, or
  authoritative state revision changes;
- shares one command-state implementation across content providers without
  replaying a selection merely because confirmation was delayed; and
- accepts the firmware's active-source `STOP_STATE` as a confirmed pause while
  continuing to reject standby and an empty source.

The branch contains five reviewable commits directly on current `origin/main`
at `8aa2db1`. The feature diff therefore contains no synthetic prerequisite or
integration-only commit.

## Linked issue

None.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

- [ ] `make check` passes (the Docker-based HTTP integration target was not run)
- [ ] Tested against a real SoundTouch device
- [x] `node --test pkg/service/soundtouchweb/frontend_test/*.test.mjs`
- [x] `go test -race -count=1 ./pkg/service/soundtouchweb`
- [x] `go vet ./pkg/service/soundtouchweb`
- [x] `go test -tags browsertest -count=1 ./pkg/service/soundtouchweb`

On exact head `b487736`, all 28 frontend module tests and the complete tagged
Chromium suite pass, including one-write, endpoint-selection, absolute-delay,
target-fencing, provider/library, and active-source pause regressions. The
changed Go package also passes under the race detector and `go vet`. No
real-device result is claimed by this PR.

The clean local feature range is `8aa2db1..b487736`.

## Checklist

- [x] No CLI or documentation update is required for this player-state correction

### DIFF
--- a/cmd/soundtouch-cli/cmd_playback_capabilities.go
+++ b/cmd/soundtouch-cli/cmd_playback_capabilities.go
@@ -1,0 +1,196 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"text/tabwriter"
+	"time"
+
+	"github.com/gesellix/bose-soundtouch/pkg/models"
+	"github.com/urfave/cli/v2"
+)
+
+// The player has to decide, per source, whether a track skip can be verified
+// at all: it confirms a skip by watching `trackID` change, because track and
+// stationName move on their own while a live stream rewrites its metadata.
+// Sources with no track concept report no trackID, and there nothing can be
+// verified.
+//
+// Which sources report what is a question only hardware answers, and answering
+// it means walking every source on a real speaker. This command exists to make
+// that walk one pass instead of one `play now --verbose` per source and a
+// manual diff: it prints one row per observation and, with --watch, a new row
+// whenever the speaker reports something different, so switching sources on
+// the speaker (or in the app) fills in the matrix as you go.
+//
+// See docs/content/docs/reference/PLAYER-SOURCE-BEHAVIOUR.md, "Track skips:
+// what can be verified", for the table these rows are collected into.
+
+// capabilityObservation is one speaker answer, reduced to the fields that
+// decide whether a skip is verifiable.
+type capabilityObservation struct {
+	Source        string
+	SourceAccount string
+	PlayStatus    string
+	StreamType    string
+	TrackID       string
+	Track         string
+	CanSkip       bool
+	CanSkipPrev   bool
+	SeekSupported bool
+}
+
+func observeCapabilities(nowPlaying *models.NowPlaying) capabilityObservation {
+	if nowPlaying == nil {
+		return capabilityObservation{}
+	}
+
+	return capabilityObservation{
+		Source:        nowPlaying.Source,
+		SourceAccount: nowPlaying.SourceAccount,
+		PlayStatus:    nowPlaying.PlayStatus.String(),
+		StreamType:    nowPlaying.StreamType,
+		TrackID:       nowPlaying.TrackID,
+		Track:         nowPlaying.Track,
+		CanSkip:       nowPlaying.CanSkip(),
+		CanSkipPrev:   nowPlaying.CanSkipPrevious(),
+		SeekSupported: nowPlaying.IsSeekSupported(),
+	}
+}
+
+// skipVerdict states, for this observation, what the player can do with a
+// skip command against this source. It is the whole point of the sweep.
+func (o capabilityObservation) skipVerdict() string {
+	switch {
+	case o.Source == "" || o.Source == "STANDBY":
+		return "n/a (nothing playing)"
+	case o.TrackID != "":
+		return "verifiable by trackID"
+	case o.CanSkip || o.CanSkipPrev:
+		return "claims skip, no trackID"
+	default:
+		return "no track concept"
+	}
+}
+
+// reportsSameCapabilities compares only what the matrix records, so a row is
+// not reprinted every poll while a track's elapsed time ticks along. A new
+// trackID is a change: that is what a skip is expected to produce.
+func reportsSameCapabilities(previous, current capabilityObservation) bool {
+	return previous == current
+}
+
+func placeholder(value string) string {
+	if strings.TrimSpace(value) == "" {
+		return "-"
+	}
+
+	return value
+}
+
+func capabilityHeader() []string {
+	return []string{"TIME", "SOURCE", "ACCOUNT", "STATUS", "STREAMTYPE", "TRACKID", "SKIP", "SKIPPREV", "SEEK", "SKIP VERDICT"}
+}
+
+func capabilityRow(observed time.Time, o capabilityObservation) []string {
+	return []string{
+		observed.Format("15:04:05"),
+		placeholder(o.Source),
+		placeholder(o.SourceAccount),
+		placeholder(o.PlayStatus),
+		placeholder(o.StreamType),
+		placeholder(o.TrackID),
+		fmt.Sprintf("%t", o.CanSkip),
+		fmt.Sprintf("%t", o.CanSkipPrev),
+		fmt.Sprintf("%t", o.SeekSupported),
+		o.skipVerdict(),
+	}
+}
+
+// playbackCapabilities polls now-playing and reports the skip-relevant
+// capabilities, once or until the watch window closes.
+func playbackCapabilities(c *cli.Context) error {
+	clientConfig := GetClientConfig(c)
+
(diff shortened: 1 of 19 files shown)


gesellix:
Thanks for this. The core mechanism is right, and it is the part that is hard to get right: one write, generation-fenced results, absolute readback delays computed from the press rather than from the write's return, and a target identity fence so a delayed result cannot land on a device you have since switched away from. Generalising that out of `Sources.js` is the correct direction.

Reviewing it turned up a handful of behaviours that did not survive the generalisation, most of them ones `Sources.js` carries explicit comments for because they only show up against real speakers. Rather than send you back a list, I rebased the branch onto current `main` and added thirteen commits on top of yours: eleven fixes, one per concern, each with a browser-level regression test that fails without it, plus two that add the tooling used to measure the speaker's skip behaviour and record what it reported. Your five commits are untouched. Happy to drop or change any of them.

## What I changed

**`unverified()` retracted a confirmation that had already happened.** It replaced whatever outcome the command had reached, so a `nowPlayingUpdated` push that confirmed a pause at t=1s was retracted by a readback that failed at t=10s: "Pause command unverified" and an amber border on a command the speaker demonstrably performed. Now it settles through a functional update, the way `settleUnverified` in `Sources.js` does.

**Nothing settled early when the event stream was live, so every press locked the UI for the whole window.** With `[2000, 5000, 10000]`, one transport press disabled Controls, Presets and Recents for ten seconds even when the speaker confirmed in 200ms, and issued three now-playing polls the WebSocket had already answered. A matching readback now settles the command when `webSocketConnected` is true, as in `Sources.js`; without an event stream the readbacks are still the only watcher and run to the end.

**`success:false` was treated as definitive.** Only a 4xx is proof the command never reached the speaker: the service produces it before any speaker call, while `sendControlResponse` maps any speaker-call error to 500, and a request that timed out after the speaker already switched looks exactly like one it never received. The hook's writes now go through checked `api` variants (`keyChecked`, `powerChecked`, `controlChecked`, and so on, beside the existing `playChecked`), and only a definitive refusal abandons verification.

Worth noting for the fixtures: `sendError` is the only path that sets `success:false`, and it always sets a status code, so a 200 carrying `success:false` does not occur in production. The rejected-write fixture now answers 400, with a 5xx case beside it that the readbacks still confirm.

**`play` never accepted `BUFFERING_STATE`.** It is a real firmware value (`pkg/models/nowplaying.go`, already special-cased in `NowPlaying.js`), and internet radio sits in it for seconds at a time, sometimes past the last deadline. Pressing play on a buffering TuneIn stream reported "unverified" on a command that was working. Buffering now confirms, for a real source only.

**Track identity was unreliable in both directions.** `playbackIdentity` folded `Track` and `StationName` into the identity, so on a live stream, whose title metadata rolls over by itself, a Next that did nothing was confirmed anyway; and on AUX, PRODUCT or radio, where nothing ever changes, a Next that worked ended unverified after ten seconds of dead transport.

The speaker does expose a usable signal for this: `trackID`. It is a stable per-track identity (`models.NowPlaying` already parses it, and the captured firmware responses carry `spotify:track:...`), so it does not move when a stream rewrites its own title, and it does change on a real skip. Skips are now verified against `trackID`, which fixes the false-confirm half properly. A definitive refusal still fails, an ambiguous write failure settles as unverified, and an error-source transition during a skip counts as a failure again.

A `trackID` alone does not cover every source, which a sweep across a real speaker made clear: `STORED_MUSIC` skips perfectly well, claims both `skipEnabled` and `skipPreviousEnabled`, and reports no `trackID` at all, while `TUNEIN` and `RADIO_BROWSER` report neither a `trackID` nor any skip claim. So the claim is what separates the two cases, and the rule is three-way:

| What the speaker reports        | Player behaviour                  |
|---------------------------------|-----------------------------------|
| a `trackID`                     | verify the skip against it        |
| no `trackID`, but `skipEnabled` | verify against the track metadata |
| neither                         | settle on the accepted write      |

The metadata fallback is safe exactly where the skip claim is present: a library changes its title only when the track really changes, whereas live radio rewrites its title with the same stream playing and claims no skip support. That is `models.CanSkip` / `CanSkipPrevious` finally earning their keep. Where neither signal exists, the command settles on a write the speaker accepted, schedules no readbacks and frees the transport immediately, rather than spending the window to end "unverified".

The same flags are deliberately *not* used to enable or disable the buttons: they flap while a source buffers. Mid-buffer the same Spotify track reported `skipPreviousEnabled` false, and in another buffering window reported no `trackID` at all, so buttons driven by them would flicker.

**Previous-track restarts before it steps back.** Measured on a SoundTouch 10: the first `PREV_TRACK` restarts whatever is playing (position 97s to 0, same track, same queue offset) and only a second press moves back; from a track that has just begun, one press steps back, and a walk down a whole album moved one track per press from 10 to 01. That first press changes no track, no `trackID` and no metadata, so every identity-based rule reported "unverified" for a command doing exactly what the firmware intends. `previous-track` now also accepts a play position that has moved backwards.

Collecting all of this is a new CLI command on the branch: `soundtouch-cli play capabilities --watch` prints a row per source with `trackID`, `skipEnabled`, `skipPreviousEnabled`, `streamType` and a verdict, and a new row whenever the report changes, and `--probe-skip` sends one `NEXT_TRACK` and reports whether the `trackID` moved, only the title moved, or nothing changed. The measurements, and what is still open (`BLUETOOTH` and `LOCAL_INTERNET_RADIO` unobserved; other media servers may report a `trackID`), are recorded in `docs/content/docs/reference/PLAYER-SOURCE-BEHAVIOUR.md` under "Track skips: what can be verified".

**Revision comparison ignored `epoch`.** `app.js acceptsNewerStatus` is deliberately epoch-aware, because a device id backed by a new `DeviceConnection`, or a restarted service, restarts its revisions at 0. If the connection was re-established mid-command, readbacks came back at revision 3 in a new epoch and the command could never confirm. `commandRevision` now carries the epoch and a single `isNewerRevision` helper applies the `app.js` rule everywhere the hook compares.

**Offline speakers lost their controls.** `disabled=${commandBusy || !source}` and the `!np` variants in `Controls.js` meant a speaker that is offline or has not been polled yet had no working power button and no working transport, which is exactly the speaker you would want to power cycle. Without state there is nothing to reconcile, so those fall back to the pre-reconciliation behaviour (send the key, report nothing) instead of going dead.

**The empty-slot preset save was gated on the command.** Clicking an empty slot saves what is playing, which is not a playback command, and the sibling star button that saves the same thing stayed enabled anyway.

## Left open

`Sources.js` has the same epoch gap in its own command path. I left it out of this branch rather than widen the diff.

`Sources` also keeps its own `commandRef` and busy state and is not gated by `commandBusy`, nor the hook by Sources' busy, so a source selection and a preset selection can run at once and their two banners can contradict each other. Not a blocker, but the two machines will want merging eventually.

## Checked and fine

Mentioning these so nobody "fixes" them later by mistake:

- `response.data.info.device_id` genuinely is present on the now-playing readback (`HandleDeviceNowPlaying` to `deviceViewForID` to `deviceView.Info`, with `models.DeviceInfo.DeviceID` tagged `json:"device_id"`), so the target fence does not misfire.
- Effect ordering in `ContentPlaybackCommand` is correct: the hook's `[deviceId]` reset effect is registered first, and `observedCommand` suppresses a spurious `onClear` on mount.
- The `TuneInBrowser` and `RadioBrowser` `invoke` closures capture the pre-clear `pendingPlay` binding, so `containerArt: pendingPlay.image` is not a null deref.
- `HandlePlayURL` returns `Source`, `Location` and `ItemName` from the same struct that was sent to `/select`, and the new `handler_test.go` asserts exactly that, so `expectedFromResponse` in `PlayURL.js` is fed the real identity.
- `Math.max(0, delay - (Date.now() - startedAt))` with the timers scheduled before the write does deliver the absolute delays you describe.

Verification on the branch: `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./pkg/service/soundtouchweb`, `golangci-lint` (0 issues), the 28 frontend module tests, and the full `browsertest` suite. The skip semantics above were measured against a real SoundTouch 10; the Player UI itself has not yet had a full hardware pass.

PR #712, opened 2026-09-12
feat(player): project logical multiroom zones

Mr-Tao:
## Summary

The embedded player currently renders every physical speaker as an independent card even when firmware reports one multiroom zone. This makes an active group look like unrelated devices and exposes both halves of a stereo pair as separate members.

This change adds a read-only logical projection layer:

- renders an authoritative multiroom zone summary on its master while keeping every logical member individually addressable for controls and device API reads;
- represents a confirmed stereo pair as one logical member without losing its LEFT and RIGHT physical identities;
- distinguishes logical-member and physical-speaker counts, including degraded availability;
- adds collapsible zone-member detail with name, model, connectivity, full IP address, physical identity, and stereo role;
- keeps current topology generation-authoritative across REST refreshes and WebSocket events, with fail-open physical cards when authority is missing or conflicting; and
- adds deterministic frontend projection, ordering, stale-response, dissolved-zone, master-handoff, member-reachability, and rendering coverage needed to maintain that boundary.

The review follow-up also clears a former master's stale cached claim after a verified handoff and preserves a genuine slave's fresh REST role, including its Leave-zone affordance, when no master projection is cached.

This layer deliberately does not add zone mutations, group-volume behavior, device settings, or stereo-pair lifecycle operations.

It is exact commit `15a97c6`, directly based on upstream `main` at `c2edd90`. The separately prepared zone mutation, member-navigation, and volume changes are not part of this PR; #714 will be restacked directly on this corrected commit.

## Linked issue

Refs #252

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

On exact commit `15a97c6`:

- `node --test pkg/service/soundtouchweb/frontend_test` (53 tests)
- `go test ./pkg/service/soundtouchweb/... -count=1`
- `go test -race ./pkg/service/soundtouchweb/... -count=1`
- `go vet ./pkg/service/soundtouchweb/...`
- `go test -tags browsertest ./pkg/service/soundtouchweb -run TestOpenZoneDetailTracksExternalTopologyChanges -count=1`
- `golangci-lint v2.13.1 run` (0 issues)
- `git diff --check origin/main...HEAD`

This is exact-source synthetic evidence. Earlier integrated UDM canaries verified a logical zone containing a stereo pair, member availability, and touch layouts, but this focused restack has not itself been deployed. The real unavailable-member canary remains outstanding.

## Checklist


gesellix:
Thanks for this. The generation and ordering machinery on the Go side is
careful work, the fail-open conflict detection does what it says, and the whole
suite is green on the branch here too: `go test -race
./pkg/service/soundtouchweb/...`, the node tests, and the browser tests.

What I would like to settle before merging is the reachability contract, not
the projection logic itself. Three findings, each reproduced against the
branch.

## 1. A non-master zone member becomes unreachable in the player

`pkg/service/soundtouchweb/device_projection.go:305` deletes every non-master
logical member from the snapshot:

```go
for _, logicalID := range candidate.logicalMembers {
    if logicalID != candidate.masterControlID {
        delete(devices, logicalID)
    }
}
```

While the zone exists, that speaker has no card at all: no volume, presets,
source selection, bass or balance, no settings, no stereo-pair create or
dissolve, and no "Leave zone". The zone detail is read-only by design in this
PR, so there is no replacement route. The only ways out are the master's zone
management or the Bose app.

Two consequences beyond the missing card:

- `static/js/app.js:447` force-clears `selectedId` when the selected device
  leaves the map, so joining a zone (from the Bose app, a speaker button, or
  this UI) kicks the user out of that speaker's detail view mid-session.
- `HandleAPIDevice` (`handler.go:493`) and `HandleDevicePowerStatus`
  (`handler.go:454`) both resolve through `deviceViewForID`, so
  `GET /api/control/devices/{member}` and `/power-status` now answer 404 for a
  zone member, while volume, key, action, zone/leave and stereo-pair on the
  same `{id}` still work. That split is going to be confusing for anyone
  driving the API directly.

Before this change a zone member was a fully controllable card. Either the
zone card needs member navigation, which is exactly what
[PR #714](https://github.com/gesellix/Bose-SoundTouch/pull/714) adds, or
members should stay in the list. See the note at the end about landing the two
together.

## 2. A master's cached zone is never cleared after a master handoff

`pkg/service/soundtouchweb/webtypes/types.go:994`: `ApplyPolledZone` rejects a
response whose `master` is not the queried device and returns `false` without
advancing `zoneAppliedGeneration` or clearing the cache. Only a standalone or
self-mastered response can clear it.

So once speaker A has cached a zone it mastered and then becomes a slave of a
different master B, which needs nothing more exotic than building a zone with B
as master from the Bose app, A's `/getZone` forever reports `master=B` and is
forever ignored. A keeps advertising the stale `master=A` topology.

Reproduced at two levels:

- After five routine polls plus a `BeginZoneEventRefresh` and apply, all
  reporting `master=B`, `conn.Status().Zone` was still
  `{Master: A, Members: [A, C]}`.
- With A stale and B fresh, `projectZoneViews` sees two competing claims and
  skips both candidates. The result is three unrelated physical cards and the
  real zone B never drawn, permanently. The fail-open behaves as designed; its
  input is what is wrong.

`TestZoneCacheRequiresAuthoritativeMaster` covers "a member's response must not
overwrite the master's". The missing half is "the queried device is no longer
the master", which should probably clear the cache rather than ignore the
response.

## 3. A visible member always renders "Standalone", so "Leave zone" disappears

`static/js/zonePresentation.mjs:97`: `effectiveZoneDetail` falls back to a
hard-coded `isStandalone: true` whenever there is no projection and
`topLevelDevice.status?.zone == null`.

A slave structurally never caches `status.zone`, because `ApplyPolledZone`
accepts master-confirmed responses only. So whenever a member's card is
visible, which is precisely the degraded cases (projection failed open per
finding 2, the master is not registered in this player, or the master has not
polled `/getZone` yet after an externally created zone), the Zone panel says
"Standalone" and offers "Add to zone", even though
`GET /api/control/devices/{id}/zone` correctly returns `isSlave: true` with a
master name. "Leave zone" is missing exactly when it is the only way out.

Confirmed directly:
`effectiveZoneDetail({isSlave: true, masterIp: '192.0.2.10', ...}, undefined,
{status: {zone: null}}, '192.0.2.20')` returns
`{isStandalone: true, isSlave: false}`.

The existing test "standalone inventory projection replaces stale REST zone
management" encodes this for the dissolve case, so the behaviour is deliberate
there. The problem is that `status.zone == null` cannot distinguish "master
whose zone was dissolved" from "slave, which never has one". A corollary worth
noting: Zone.js's `isSlave` branch is now effectively dead code, since it needs
`selected` to be present in the projection, and such a device has already been
deleted from `devices` and auto-deselected.

## Test coverage

Three gaps line up with the findings above:

- Nothing covers the master-handoff clear path from finding 2. The nearest
  test asserts the opposite half of the rule.
- Nothing asserts that a genuine slave (a device with `status.zone == null`)
  keeps its member and leave affordances, which is the gap that produced
  finding 3.
- The PR is candid that the unavailable-member canary was never run on
  hardware. All three findings bite hardest in the partially offline path, so
  that gap matters more than usual here.

## Smaller notes, not blocking

- `websocket.go:1015` adds an unconditional serial `GetZone()` to every poll
  round for every device, including models that previously skipped the group
  fetch. Correct, since all models answer `/getZone`, but it lengthens each
  cycle and adds another full client timeout for an unreachable speaker.
- `websocket.go:798`: `OnZoneUpdated` is the only event handler that does not
  call `MarkEventStreamActivity`, so a speaker whose stream carries only zone
  events looks idler than it is.
- `handler.go:1059`: `deviceViewSnapshot()` now runs once per WebSocket client
  rather than once per broadcast. The comment explains why and it is correct,
  but it is O(clients x devices) per broadcast, and broadcasts now also fire on
  every zone poll that changes something.
- The "Docs or CLI help updated" checkbox is unticked for a user-visible
  feature.

No convention problems: no real IPs, MACs or device IDs anywhere in the diff,
and the `ETag` literal is untouched.

## How I would like to land this

**As a stack, with 714 on top of 712.** On its own, this PR takes away every
control for a zone member and gives nothing back, so a release containing only
712 would be a regression for anyone who uses zones. With
[PR #714](https://github.com/gesellix/Bose-SoundTouch/pull/714) on top, the
member is reachable again and finding 1 turns into a design decision rather
than a hole. 714 has one finding of its own, which I have left on that PR.

**Please also rebase onto current main.** This branch is based on a commit from
before v0.132.0, which has since shipped, and it touches files that moved in
the meantime (the player's preset star was extracted into a shared component,
and the now-playing card and Library rows changed with it). GitHub reports no
conflicts, but a merge does not re-run the suite against today's main, and I
would rather see a fresh CI run on the rebased branch than trust a clean merge
result. Rebasing 712 first and then re-stacking 714 keeps the stack honest.

PR #714, opened 2026-09-12
feat(player): open zone member device details

Mr-Tao:
## Summary

Depends on #712. The child-specific diff is `89c5b66..60688ef`.

The embedded player projects a multiroom zone as one logical card, but its
expanded member rows are currently read-only. Because physical zone members
are intentionally absent from the top-level device list, there is no route to
their full controls without first dissolving the zone. In particular, an ST10
inside a zone cannot reach the existing stereo-pair creation workflow.

This change makes each logical member header a device-detail action:

- loads the physical member through the existing device endpoint without
  replacing the authoritative logical-device projection;
- preserves logical stereo-pair identity while using its firmware master as
  the control target;
- shows eligible and ineligible stereo-pair presentation through the existing
  lifecycle UI;
- returns to the originating zone and never issues a zone mutation merely to
  inspect a member;
- fences late member-detail responses after back or other navigation;
- retains a newer member status event when the initial REST detail response
  arrives later; and
- remounts device-scoped child state when the selected target changes, so a
  delayed zone or recents response cannot appear under the wrong speaker.

This is exact commit `60688ef`, directly based on the logical-zone projection
in #712 at `89c5b66`. The stereo-pair lifecycle it exposes is already present in
upstream `main`. It does not change pairing or multiroom mutation semantics.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

On exact commit `60688ef`:

- `node --test pkg/service/soundtouchweb/frontend_test/*.test.mjs` (49 tests);
- `go test -count=1 ./pkg/service/soundtouchweb`;
- `go test -tags browsertest -count=1 ./pkg/service/soundtouchweb`;
- `go test -tags browsertest -count=1 -run '^TestZoneMember' ./pkg/service/soundtouchweb`;
- `go test -race -count=1 ./pkg/service/soundtouchweb/...`;
- `go vet ./pkg/service/soundtouchweb/...`;
- `golangci-lint v2.13.2 run --timeout=5m`;
- the focused browser scenario uses the actual zone and stereo projection to
  open an eligible ST10, an ineligible ST20, and a logical stereo pair, then
  returns to the original zone after each;
- the same scenario records zero zone-mutation requests and verifies that a
  newer status event survives a delayed REST detail response;
- `node --check pkg/service/soundtouchweb/static/js/app.js`; and
- `git diff --check 89c5b66..60688ef`.

This focused restack has not been deployed as a separate build. The previously
deployed integrated player established the real logical-zone and stereo-pair
topology, but touch-browser acceptance of this exact navigation remains to be
performed before claiming physical verification.

## Checklist


gesellix:
Thanks, this is the missing half of
[PR #712](https://github.com/gesellix/Bose-SoundTouch/pull/712): that one makes
a zone member disappear from the device list, and this one gives it a route
back. Reviewed as `git diff pr712 pr714`, so the notes below are about this
PR's own change. Go tests pass on the branch.

Two findings, both reproduced in headless Chrome using the PR's own fixtures.

## 1. Member to sibling member loses `stereoPair`, and offers "Continue cleanup" on a healthy pair

`static/js/app.js:526-553`:

```js
const logicalMember = devices[originId]?.zone?.members?.find(...)
...
device = {
    ...resp.data,
    info: {
        ...resp.data.info,
        name: logicalMember?.name || resp.data.info?.name,
        type: logicalMember?.type || resp.data.info?.type,
    },
    stereoPair: logicalMember?.stereoPair,
};
```

`name` and `type` fall back to the response, but `stereoPair` does not: it is
taken only from the origin's zone projection, discarding
`resp.data.stereoPair`, which `deviceViewForID` does return for a pair master.

When `originId` is itself a collapsed zone member, `devices[originId]` is
`undefined`, which is the entire premise of the stack, so `logicalMember` is
`undefined` and `device.stereoPair` is set to `undefined`.

That path is reachable: `HandleGetZone` (`handler.go:1277-1291`) returns the
full projected member list for a slave too, and `effectiveZoneDetail`
(`static/js/zonePresentation.mjs:70-90`) returns every member except the
master regardless of which member is selected. So a member's own page renders
clickable rows for all of its siblings.

On the sibling's page:

- `DeviceDetail` (`app.js:361`) sees `!device.stereoPair`, so the **"Remove
  from AfterTouch"** button reappears for a stereo-pair master, which 712
  deliberately hides.
- `StereoPair.js:57` has `pair = device?.stereoPair` undefined, so
  `recoveryGroup = !pair && hasConfiguredGroup(state.group)` becomes true (the
  real `GET /stereo-pair` does return `group`, see `handler_stereopair.go:39`).
  A healthy pair renders as **"Recovery required" / "Continue cleanup"**, and
  that button calls `stereoPairDissolve`.

Reproduced with chromedp against the PR's `newZoneMemberNavigationApp`
fixtures:

```
DIRECT master->pair:  removeButton=false recoveryUI=false
NESTED member->pair:  removeButton=true  recoveryUI=true  pairMemberRows=4
```

Suggested fix: `stereoPair: logicalMember?.stereoPair ?? resp.data.stereoPair`.

Worth noting the same origin-dependency fires at first order too, whenever the
master's `devices[originId].zone` projection has not arrived yet while the REST
zone detail has: Zone.js renders rows purely from the REST response in that
window.

## 2. Clicking the row of the device you are already on swallows a Back press and freezes the view

`Zone.js:151-154` wraps every logical row in the open button, including the
master row and the member's own row. On the master's own page, clicking its
row calls `openZoneMember(masterId)` with `originId === controlId ===
selectedId`. The page does not change, but `selectedMember` now holds a frozen
snapshot, and `backFromDevice` (`app.js:563-573`) then navigates to the page
you are already on:

```
afterSelfClick="Atrium192.0.2.10"  afterBack="Atrium192.0.2.10"  afterBack2="Devices"
```

While `selectedMember` is set, `selectedDevice` (`app.js:395-397`) prefers the
frozen snapshot over the live `devices[selectedId]`. `mergeSelectedMemberStatus`
keeps `status` fresh, but `stereoPair` and `zone` stay pinned at click time, so
a pair created or dissolved elsewhere no longer shows on that page. The same
staleness applies to any member that later becomes visible in `devices`, for
instance when the zone is dissolved while you are looking at it.

Suggested fix: skip when `controlId === selectedId`, and prefer the live
`devices[selectedId]` entry whenever one exists.

## 3. Back from a nested member lands on the device list, not the zone

`app.js:563-573`: the `originId` of a nested member is another collapsed
member, which is never in `devices`, so Back falls through to
`navigate('devices')`. A single-entry origin is not enough for a list that is
browsable transitively.

## 4. `GET /api/control/devices/{id}` quietly stopped reporting `zone`

`device_projection.go:123-130`: `deviceViewForID` now calls
`projectLogicalDeviceEntries` directly, and only `projectZoneViews` populates
`deviceView.Zone`. So the detail endpoint, and `/now-playing` with it, no
longer report `zone` for a zone **master** either. No frontend consumer breaks
today (Zone.js reads `devices[...]?.zone` from the list and WS frames, and
`discreteCommand.js:408` uses only `status` and `info`), which is why nothing
fails, but it is a behaviour change from 712 that the new doc comment does not
mention. If the intent was only "do not hide zone members", then
`projectCapturedDeviceEntries(...)[id]` plus a zone-member fallback would
preserve `zone`.

## Test coverage

`zone_member_navigation_browser_test.go` walks master to member and back three
times. Not covered:

- member to sibling member, and the stereo-pair and remove-button state that
  results (finding 1).
- clicking the row of the currently open device (finding 2).
- the "Leave zone" action from a member page. That a member is otherwise
  un-leavable is the stated reason for the feature, and no test exercises it.
- failure paths: `api.device` answering 404 or throwing, and the "Device not
  found" render. An unregistered zone member (where `controlId` falls back to
  `zoneMemberIPs[deviceID]`) always 404s, yet its row is still clickable.
- the `selectedMember?.controlId !== selectedId` guard in the redirect effect
  (`app.js:502-507`), which is what stops the app bouncing back to the device
  list.
- `mergeSelectedMemberStatus` has a dead end: if the fetch failed
  (`device: null, loading: false`), a later `status_update` only sets
  `pendingStatus` and the page stays on "Device not found" indefinitely.

The Go-side changes are covered by `zone_member_navigation_test.go`, and the
pre-existing `TestHandleAPIDevice_HiddenPairMemberNotFound` still guards the
hidden-peer 404.

No convention problems: RFC-5737 addresses and placeholder device IDs
throughout, `ETag` untouched.

## Landing this

Please rebase [PR #712](https://github.com/gesellix/Bose-SoundTouch/pull/712)
onto current main first (v0.132.0 has shipped, and the player files this stack
touches moved with it), then re-stack this branch on top of it, so CI runs
fresh on both. I would like the two to land together: 712 alone removes every
control for a zone member, and this PR is what gives them back.

gesellix:
I rebased this onto current main and pushed two commits on top of yours. Your
commit is unchanged apart from the rebase.

The premise shifted underneath this PR: since the update to
[PR #712](https://github.com/gesellix/Bose-SoundTouch/pull/712), zone members
stay in the device list as their own cards (they now carry an "In group ·
<master>" badge, from [PR #748](https://github.com/gesellix/Bose-SoundTouch/pull/748)).
So a member no longer has to be reached around the inventory, and the parts
built for that started to work against it:

- **`deviceViewForID` bypass** (`eab53545`): skipping the zone projection
  meant the detail API stopped reporting a master's `zone` and a member's
  `zoneMembership`. It goes through the normal snapshot again, and the Go
  test now checks that the detail API reports both roles.
- **`selectedMember` snapshot** (`f1cb36dd`): with members in the inventory,
  every click stored a copy of an entry the inventory already keeps current.
  Only its status was merged back, so a pair or zone change made elsewhere
  never reached the open page. A member is now opened by selecting its live
  entry, which also removes `mergeSelectedMemberStatus` and
  `completeSelectedMemberLoad` together with their browser test.

The same commit covers the other two points from my first review: Back now
keeps a stack of origins, so it retraces member-to-member navigation
(skipping pages whose device has gone), and a member row is a button only
when it leads somewhere, so the row of the page you are on no longer freezes
the view.

What stays is yours: the clickable member rows and their styling, Back
returning to the zone you came from, the per-target remount of the detail
page, and your master-to-member browser test. A new browser test walks
master, member, sibling pair and Back again.

Verified on the pushed head: `go test ./...`, race tests, 57 node tests,
the browser tests and `golangci-lint` are green. Touch-browser acceptance on
real hardware is still open, as you noted.

PR #722, opened 2026-09-12
feat(stereo): harden same-zone pair transitions

Mr-Tao:
## Summary

Stereo-pair creation currently requires both candidates to be standalone. The
SoundTouch firmware can also transition two speakers that already belong to
the same temporary multi-room zone, without removing and re-adding them in the
application first.

This change makes that transition explicit and verifiable:

- accepts two standalone candidates or two candidates whose fresh zone views
  describe the same temporary zone and contain both device IDs, including a
  larger zone with a third master;
- repeats group and zone validation after the persistence preflight so stale
  topology cannot be mutated;
- verifies both the resulting stereo group and standalone zone state after a
  successful transition;
- on partial failure, requires compensation to restore both empty group state
  and the original zone topology; and
- preserves an otherwise verified stereo pair as degraded when only its zone
  readback is unavailable, rather than destructively rolling it back.

It preserves the reviewed #655 behavior that speakers may use different Marge
accounts as long as they use the same backend.

The PR is exact commit `aeb3a16`, directly based on upstream `main` at
`b6fe8f9`, which already contains #655.
It changes only the stereo coordinator, its focused tests, and the matching CLI
documentation. It does not include player-zone UI, volume, settings, event, or
browser changes.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

## How was it tested?

On exact commit `aeb3a16`:

- `go test ./pkg/stereopair` passes;
- all create-path tests pass under the race detector;
- `go vet ./pkg/stereopair`, `gofmt`, and `git diff --check` pass; and
- focused regressions cover candidates in a larger existing zone, topology
  drift before mutation, verified post-create zone state, exact compensation,
  and degraded outcomes when compensation cannot be proven.

The integrated canary previously exercised the firmware transition on real
SoundTouch 10 speakers. This focused restack has not repeated that physical
transition with a retained before/after capture, so the PR does not claim new
standalone hardware evidence.

## Checklist


gesellix:
Thanks for this. The rejection paths are thorough (six malformed or
inconsistent zone-view cases, one-sided zones, differing zones, post-preflight
drift), and I am glad the tests exercise compensation rather than only the
happy path. `go test ./pkg/stereopair/...` passes on the branch.

Two findings I would want fixed before this lands, both reproduced with probe
tests against the branch.

## 1. A rolled-back same-zone pairing leaves a stale generation and blocks all future pairing

`verifyCompensation` via `verifyCompensationZone`
(`pkg/stereopair/coordinator.go:1069`) compares the post-compensation zone
against `states[i].zone`, which is the **pre-pairing** zone read in
`preflightCreate` (`coordinator.go:1280`).

In the same-zone flow that pre-pairing zone is the temporary zone, and this
PR's own premise is that the firmware dissolves it during the transition:
`verifyCreatedPairZones` (`coordinator.go:1537`) asserts the speakers are
standalone afterwards. Compensation only calls `RemoveGroup()`, and nothing
recreates the temporary zone, so the zone assertion can never pass for the
mutated member.

The consequence is in `coordinator.go:753-762`:

```go
result.CompensationComplete = compensationVerified(states)
if !result.CompensationComplete {
    ...
    return
}

for _, ref := range compensationGenerations(states) {
    if err := c.cleanupGeneration(result, ref); err != nil {
```

`compensationVerified()` is false, so the function returns before the
generation cleanup loop. The physical group was removed successfully, but the
Marge generation created by the master is never retired. The next `Create`
then hits `generationPreflight` / `EnsureMargeNoGroupGenerations` and refuses
to mutate ("persisted group generation X still contains device Y"). The user
is left with physically clean speakers that cannot be paired again without
backend surgery.

Probe output on the branch (LEFT add ok, RIGHT add fails, LEFT rolled back and
standalone):

```
status=degraded compensationComplete=false removeCalls(L)=1 group(L)={ID: ... empty}
member0: verified=true compVerified=false compErr=compensation did not restore the pre-pairing zone topology
generationCleanup calls=0
```

The LEFT group reads back empty, so the rollback fully succeeded, and cleanup
still did not run. Before this PR the same scenario cleaned up, which the
untouched `TestCreatePartialFailureIsCompensatedAndReported` still asserts.

`TestCreateSameZonePartialFailureRequiresZoneTopologyRestoration`
(`coordinator_test.go:449`) locks the new behaviour in: it asserts
`CompensationComplete == false` and a non-nil `CompensationError` for exactly
this case, and says nothing about the generation, so the real damage is
invisible in the suite. Every same-zone compensation test uses `New(...)`
without generation persistence, which is why `cleanupGeneration` is a no-op
throughout.

Three ways out, in my order of preference: have compensation actually restore
the zone (`setZone`), or make the zone postcondition "standalone or the
pre-pairing topology", or at the very least stop gating generation cleanup on
a postcondition the coordinator never tries to satisfy.

Related, same mechanism: `verifyCompensation` loops over all states, including
the member `compensateCreate` skipped (`coordinator.go:707-711`) and never
touched. If that untouched speaker's zone drifts for an unrelated reason, its
`CompensationVerified` goes false and blocks retiring the other speaker's
generation. That one affects the pre-existing standalone flow too.

## 2. One transient `/getZone` turns a verified pair into a reported failure, and the CLI then hides the group ID

`verifyCreatedPairZones` (`coordinator.go:1515-1544`) does exactly one
`GetZone()` per speaker with no settle loop, unlike every other verification in
this file, which retries through `c.uncertainOutcomeDelays`
(`coordinator.go:398-409`, `742-751`, `919-928`). A lagging zone view right
after a zone-to-stereo transition is expected firmware behaviour, and a plain
timeout is enough to trigger this.

Probe, with the group fully verified on both speakers and one transient
timeout on RIGHT's zone read:

```
status=degraded zoneCalls(R)=2 (no retry)
members verified: true true; verErr=get zone after verified group mutation: i/o timeout
compensationAttempted=false; group=PAIR-ID ... GROUP_OK
```

The pair exists and is correct. `finish()` (`coordinator.go:1835`) returns a
non-nil error for degraded, and `cmd/soundtouch-cli/cmd_group.go:62-68` takes
the `err != nil` branch: it prints "Failed to create stereo pair" and never
reaches `printGroup`. So the user is told the operation failed and is not given
the group ID they need for `rename` or `dissolve`. The web handler fares better
because `applyStereoPairProjection` publishes `member.Group` regardless.

Keeping the group on the result is right. Putting that zone read through
`uncertainOutcomeDelays`, and letting the CLI print the group ID on a
degraded-but-verified create, would finish the job. Note also that the member
results are internally contradictory in this case: `Verified == true` next to a
non-nil `VerificationError`, which `printGroupResultDetails`
(`cmd_group.go:260`) renders as "verification failed".

## 3. Two definitions of "standalone" now coexist

`parseCreateZone` (`coordinator.go:599`) treats `len(devices) == 1 && master ==
deviceID` as standalone, while `models.ZoneInfo.IsStandalone()`
(`pkg/models/zone.go:127`) uses `len(Members) == 0`, and that is what
`preflightStandaloneZone` (`coordinator.go:1293`) still uses for rename and
dissolve.

They disagree on at least two real shapes:
`<zone master="SELF"><member>SELF</member></zone>` is standalone to create but
"currently in a zone" to rename and dissolve, and `<zone master="OTHER"/>` is
standalone to rename and dissolve but a hard `ErrConflict` to create. Create
fails closed in both directions, so this is a consistency problem rather than a
data-loss one, but it means a pair that create accepts may not be renameable or
dissolvable.

## 4. Accepting a zone whose master is a third speaker (question, not a finding)

`validateCreateZones` only requires that both candidates appear in the same
zone, and
`TestCreateTransitionsCandidatesFromLargerZoneWithThirdMaster`
(`coordinator_test.go:229`) deliberately accepts two slaves of a zone mastered
by an unrelated speaker. That third device is never contacted, never re-read,
and `verifyCreatedPairZones` inspects only the two candidates, so a dangling
member reference left on the third speaker would go unnoticed.

The PR describes the use case as "two speakers that already sit in the same
temporary zone", which is narrower than what the code accepts. I could not
determine the firmware's behaviour here from the repo, so this is a question
rather than a finding: has that shape been tried on hardware? If not, I would
rather narrow it to "the zone contains exactly the two candidates" for now.

## Test coverage

Beyond the two gaps implied above (no test asserts what happens to the
persisted generation after a same-zone rollback, and none covers a transient
zone read that a retry would fix), one note on the fake: `fakeClient.GetZone`
now synthesises an error when `f.zone == nil` (`coordinator_test.go:84-86`),
and `f.zone, f.zoneErr = f.getZone(...)` persists the injected value, so once a
hook returns `(nil, err)` the zone stays nil for every later call. Harmless for
the single-pass tests as written, but it would make retry tests misleading.

No convention problems: RFC-5737 addresses and placeholder device IDs, no real
device data.

## Landing this

Please rebase onto current main when you pick this up. The branch predates
v0.132.0, and while GitHub reports no conflicts, a merge does not re-run the
suite against today's main.

gesellix:
I rebased this onto current main (clean, no conflicts) and pushed five
commits on top of yours, one per point from my earlier review. Your commit is
unchanged apart from the rebase.

- **Same-zone rollback now retires its generation** (`fix(stereo): accept a
  standalone speaker after a same-zone rollback`). Compensation accepts a
  standalone speaker as the clean end state, alongside the restored
  pre-pairing zone; any other zone still fails verification. The same-zone
  rollback test now asserts completion and the retired generation, and a new
  case covers a speaker that joined another zone. A correction to my earlier
  review: its "untouched member" sub-point was framed wrongly. In create the
  slave is always attempted after the master, so a failed `addGroup` still has
  to be proven empty, which `TestCreatePartialFailureIsCompensatedAndReported`
  rightly checks. The zone rule above covers the real case.
- **The zone check after create settles** (`fix(stereo): settle the zone
  check after a verified pair is created`). It retries through
  `uncertainOutcomeDelays` like the other verifications. The fake client no
  longer keeps a hook's error for every later read.
- **The CLI prints the pair it kept** (`fix(cli): show a created stereo pair
  even when create reports a problem`), with its ID, and describes an error on
  a verified member as a failed follow-up check rather than "verification
  failed".
- **One definition of standalone** (`fix(stereo): use create's reading of
  "standalone" for rename and dissolve`). Rename and dissolve parse the zone
  view with `parseCreateZone` too.
- **Two-speaker zones only** (`fix(stereo): accept a same-zone transition only
  for a two-speaker zone`). A zone with a third speaker, as master or member,
  is refused until that shape has been tried on hardware. The CLI reference
  says so.

One limitation worth knowing, from a hardware test on our zone work this week:
a SoundTouch 10 that was a zone member when its master was unplugged stayed
internally in its slave state, with itself as master, while `/getZone`
reported it standalone. Selections looped and `now_playing` froze until a
reboot. A same-zone transition cannot see that state through `/getZone`, so
it is not something this PR can guard against; I mention it in case you run
into it while testing.

Verified on the pushed head: `go test ./...`, race tests for the stereo,
CLI and player packages, and `golangci-lint` are green. The same-zone
transition itself is still untested on hardware here.

PR #740, opened 2026-09-13
feat(player): expose capability-driven settings API

Mr-Tao:
## Summary

The typed SoundTouch settings operations are available in the client, but the embedded player has no safe server boundary through which to use them. Directly exposing those calls would also let a stale browser mutate a different physical speaker after a stereo projection or same-address replacement changed ownership.

This change adds a capability-driven settings API to the embedded player:

- returns one fresh settings snapshot whose controls require both advertised endpoint support and usable readback;
- identifies the exact physical speaker represented by that snapshot and requires the identity on every mutation;
- revalidates the live `/info` identity before issuing a command, rejecting stale, missing, or replaced targets before touching hardware;
- validates request bodies and finite firmware values for clock display, automatic standby, language, rebroadcast latency, Bluetooth pairing, paired-device clearing, and source naming;
- verifies mutations through setting-specific readback and distinguishes confirmed, rejected, and unverified outcomes;
- preserves a targeted readback's confirmed outcome when only the subsequent complete settings refresh fails; and
- exposes whether the separately implemented Wi-Fi onboarding workflow is actually mounted, without adding network diagnostics or changing that workflow.

The PR is exact commit `bfd47e9`, based on upstream main `aabfcd2`. It builds on the client operations merged in #673 and the stereo projection already present in main, but does not change player presentation, playback, or topology.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

On exact commit `bfd47e9`:

- `go test ./pkg/service/soundtouchweb ./cmd/soundtouch-service -count=1` passes;
- `go test -race ./pkg/service/soundtouchweb -count=1` passes;
- `go vet ./pkg/service/soundtouchweb ./cmd/soundtouch-service` passes; and
- `golangci-lint v2.13.1 run`, formatting, and `git diff --check` pass.

The full `go test ./... -count=1` run passes every package except the unchanged environment-sensitive `TestCheck443Reachability_LANProbeMatchesListenerOutcome`, because LAN port 443 is reachable on this host while that test assumes it is not.

The settings endpoint shapes and alternate/restored values were previously exercised through the integrated canary on SoundTouch 10, 20, and 30 hardware. This focused restack has not been deployed as a separate build.

## Checklist


gesellix:
Thanks, this is a careful boundary. The identity handling in particular does
what the description promises: the target header is required, the per-device
lock is taken before the registry lock everywhere, the registry entry is
re-read, and the speaker's own `deviceID` is checked live before and after each
mutation. The confirmed / rejected / unverified split also reads right.

I rebased this onto current main (clean, no conflicts) and pushed three small
fixes on top of your commit, which is unchanged apart from the rebase:

- **An unreadable source list is a read error, not "not supported".** Bluetooth, pairing, paired-device
  clearing and source naming derive their support from `/sources`. When that
  read failed, the three mutations answered 409 "not supported by this device"
  to a single slow read. `requireSetting` now takes the snapshot error keys a
  setting's support depends on and answers 502 with the read error when one of
  them failed.
- **The clock-display timezone is trimmed once, right after decoding.** A
  whitespace-only `timeZone` was refused with 409 on a speaker without a
  timezone, and turned a confirmed enable change into 202 "unverified" on one
  with a timezone.
- **Source names are validated before writing.** Control characters (which the
  XML encoder replaces with U+FFFD) and names over 64 characters are a 400 now.
  The firmware's own length limit is not known; 64 is our conservative bound
  and says so in the code.

Each fix comes with tests; the first two fail without their fix. `go test ./...`, the race tests and
`golangci-lint` are green on the pushed head. The UI in
[PR #742](https://github.com/gesellix/Bose-SoundTouch/pull/742) is next; it
will need a re-stack on this.

PR #742, opened 2026-09-13
feat(player): add capability-driven settings UI

Mr-Tao:
## Summary

The capability-driven settings API has no player presentation, leaving users to call it manually even for ordinary device-local operations.

This change adds one subordinate Settings surface to device detail:

- loads settings only when the disclosure is opened and omits controls whose support or current value was not confirmed;
- presents clock display and synchronization, automatic standby, system language, rebroadcast latency, Bluetooth pairing and paired-device clearing, source naming, and the existing Wi-Fi onboarding entry point;
- binds component state, requests, and delayed responses to the exact physical speaker represented by the loaded snapshot;
- distinguishes applying, confirmed, unverified, and rejected outcomes instead of optimistically claiming success;
- retains the current command-state, presets, stereo lifecycle, AirPlay guidance, and device-removal behavior from main; and
- deliberately leaves network diagnostics and acoustic controls to their separate focused changes.

The UI is exact commit `748225c` on top of settings API commit `bfd47e9` from #740. Until #740 merges, this stacked PR includes that prerequisite as its first commit; the UI itself is the seven-file second commit and does not modify the API implementation.

## Linked issue

Refs #252

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

## How was it tested?

On exact commit `748225c`:

- all 53 Node frontend tests pass;
- all three focused tagged Chromium settings tests pass, including narrow mobile layout, keyboard focus/activation, and a real CDP touch event;
- the inherited settings API package and route tests pass; and
- `golangci-lint v2.13.1 run`, JavaScript parsing, formatting, and `git diff --check` pass.

The same settings surface was previously exercised through the integrated canary on SoundTouch 10, 20, and 30 hardware and rendered on iPad and desktop Chromium. This focused restack has not been deployed as a separate build.

## Checklist


gesellix:
Thanks. Binding the component to the snapshot's `targetIdentity` and fencing
late responses by generation is the right shape, and the narrow-layout and
touch coverage is welcome.

With [PR #740](https://github.com/gesellix/Bose-SoundTouch/pull/740) merged, I
re-stacked your UI commit onto current main (clean, no content changes) and
pushed six commits on top.

The most important one first: **the three async settings browser tests were
not checking anything past their first `await`** (`test(player): await the
async settings browser contracts`). Their scripts were passed to
`chromedp.Evaluate` without `awaitPromise`, so Evaluate returned as soon as
each script started and any rejection was lost; even a deliberate `throw`
passed. Awaited, the main contract failed at its first target change,
because of a real race in the component:

- **First load vs. the reset effect** (`fix(player): keep the first settings
  load when the reset effect runs late`). The target-reset effect also ran on
  mount, after paint, so it could orphan a load that had already started; and
  a target change with the section open never loaded the new target, since no
  toggle event follows. The reset now runs only on a real target change and
  loads the new target itself when open. In the player this was masked by
  `DeviceDetail` keying `Settings` per target.

The other findings from my review:

- **Pair settings go to the registry key** (`fix(player): address a stereo
  pair's settings by its registry key`). `controlId` came from the master
  member's `ipAddress`, which differs from the service's id for a speaker
  added by hostname or with a second interface.
- **Answerless failures are unverified** (`fix(player): show an answerless
  settings change as unverified, not rejected`). The API's own JSON answer is
  kept as is; only a transport error or a non-JSON 5xx becomes "unverified",
  and an unverified result without a snapshot re-reads the settings.
- **One message for an unverified Bluetooth clear** (`fix(player): show an
  unverified Bluetooth clear once, not also as an error`). `bluetoothClear`
  is no longer a section error key; the test's clear mock now carries it as
  the API sends it.
- **Reload** (`feat(player): let settings be reloaded once they are loaded`),
  placed at the end of the section so Tab from the header still reaches the
  first control, as your keyboard test expects.

Not done: the control snapping back to its old value while "Applying change…"
is shown, and the missing disabled style on the segmented sync control. Both
are cosmetic; happy to leave them for a follow-up.

Verified on the pushed head: `go test ./...`, race tests, 72 node tests, all
browser tests (the settings ones now actually awaited, repeated to rule out
flakiness) and `golangci-lint` are green.

